PR #22, opened 2015-04-02
added a method for line and circle

srandoux:


hzeller:
Thanks; I think it is useful to have these operations available. Some remarks below that would need to be addressed before merging.

I think these functions should go into the graphics.h file, so that they are not members of Canvas, but a utility function that draws on a Canvas (like DrawText()). Canvas is meant to be the minimum viable implementation of something 'canvasy'.

(also, obviously, they should't be called myCircle(), myLine(), but DrawCircle() and DrawLine()).

Try to adhere to the code style used in the project, like 2 character indent and open curly braces on the same line as while() or if().

The x and y coordinates should be int instead of uint8_t - it is possible to have way larger Canvases than fits into a uint8 (Look at SetPixel() - it allows for int in its x/y parameters).

Your line implementation makes use of floating point operations, which is probably not the cheapest operation on an ARM. Can you re-formulate that with fixed integer operations (or alternatively convince me, that it is as fast with floating point on the Raspberry Pi than with integer with a benchmark).

srandoux:
Hi, 
I just reused some code I had around but I agree with you. I'll do the updates and keep you posted.

Thanks,

PR #25, opened 2015-04-15
Enable error on warnings & Fix compilation errors

mtottenh:
I was attempting to cross compile this library on my Mac and compilation failed with a few trivial errors/warnings.

This PR should fix those, it's very small & self contained.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CXXFLAGS=-Wall -O3 -g
+CXXFLAGS=-Wall -O3 -g -Werror
 OBJECTS=demo-main.o minimal-example.o text-example.o led-image-viewer.o
 BINARIES=led-matrix minimal-example text-example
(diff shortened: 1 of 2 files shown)


hzeller:
Are you really cross-compiling to generate 32Bit ARM code for the Raspberry Pi or are you just compiling on that platform to see if it compiles ? It sounds like the latter as the off_t seems to be 64Bit on your platform. 
The target architecture is the Raspberry Pi and it is very specific to it, so you need to make sure to get the fully compatible cross-compile headers for it.

hzeller:
While I didn't pull this directly, I fixed things that wouldn't work with clang.
(Looks like clang on the Raspberry Pi is actually too buggy to use, but I could verify this on an amd64 machine).
The %ld format conversion was kept (the %td actually didn't work).

I'll close this pull request now; if there is anything left that needs to be addressed, please send a separate one.

PR #36, opened 2015-06-28
Adding Python bindings

Saij:
I've added some basic Python bindings for the RGB Matrix library.

Currently only the classes RGBMatrix and FrameCanvas are supported.
Also ported some of the samples to python


hzeller:
Thanks! Merged.

hzeller:
Thanks, I think many people will like your contribution!

Could you create another pull request and update the setup.py in a way that changes in the *.pyx file automatically trigger a new cythonizing to generate a new rgbmatrix.cpp ?
We still can supply the *.cpp file by default in the repository so that people don't have to install cython, but it should be easy to change.

Of course, the regular setup (when people just compile stuff) should also be resilient if cython is not installed; not sure if the setup.py will creep out if it imports a non-existing package.
So maybe that would need to be a separate setup for that and a Makefile-rule
  %.cpp : %.pyx
      $(PYTHON) cython-setup.py
.. or something ? Then it only will be triggered if the *.pyx file is newer
(I have done very little Python before, so I leave it to you to do the best way this would be typically done).

Saij:
I don't think it is a good idea.

On the cython documentation it reads:
> It is strongly recommended that you distribute the generated .c files as well as your Cython sources, so that users can install your module without needing to have Cython available.
> It is also recommended that Cython compilation not be enabled by default in the version you distribute. Even if the user has Cython installed, he/she probably doesn’t want to use it just to install your module. Also, the installed version may not be the same one you used, and may not compile your sources correctly.

So maybe it is not a good idea to implement cythonizing in the Makefile - not everyone has cython installed.
Whoever chances the .pyx file should also cythonize it and push booth - the .cpp and the .pyx file.

But if you want it I could add the lines
%.cpp: %.pyx
$(CYTHON) -cplus ... (I have no idea about the Makefile syntax ^^)

hzeller:
I am not arguing that we shouldn't include the generated file in the distribution. It should be there, because, as you said, not everybody has cython installed.

But if somebody wants to change the pyx file (or the library changes), the build instructions how to get t here should be included in the Makefile, otherwise people are stuck.

I'll add the rule to the Makefile.

Saij:
Thanks (I'm not very good with Makefiles).

hzeller:
Ok, in
https://github.com/hzeller/rpi-rgb-led-matrix/commit/48c870a0dca534ccb65687623c227e86cfaaf501
I added a generic rule to create the *.cpp file, but that file is not a direct dependency. So only someone who actively updates the *.pyx file can trigger a re-generation with
  make rgbmatrix.cpp

Saij:
Thank you very much

daveythacher:
This is a bad approach. This should be a repository of itself. That repository would then submodule this repository. Who owns the new repository is a question. The issue tracker is littered with Python junk, this would give it a dedicated place.

I think the management of the bindings is different enough as say a pure python version. The primary advantage to using a submodule here is to work with specific versions of this repository. So the Python bindings stay functional. If a new feature comes out and the Python bindings are not supporting it it will not exist till this repository has gotten around to work it out.

From what I understand the support of this repository is primarily focused on the original C/C++ base.

PR #38, opened 2015-07-01
Added define "EXPERIMENTAL_BITPLANES"

Saij:
If the new pinout is used we should be able to set the maximum available
PWM bits. Lower available PMW bits will result in better refresh rate.
Little sideffect: the smaller the value the darker the LEDs will be.

(Tested with stock Rasbian kernel - no RT patch)
Testing commandline:
    
    ./led-matrix -r 32 -P 1 -c 1 -p 4 -D 4
    ./led-matrix -r 32 -P 1 -c 1 -p 8 -D 4
    ./led-matrix -r 32 -P 1 -c 1 -p 11 -D 4

Results with EXPERIMENTAL_BITPLANES=4

    With 4 active bitplanes
        Brightness: low (but viewable) - you can see the white LED chips
        Refreshrate: ~ 450Hz
        No flickering seeable
    With 8 active bitplanes
        not possible
    With 11 active bitplanes
        not possible

Results with EXPERIMENTAL_BITPLANES=8

    With 4 active bitplanes
        Brightness: bright (you don't the white LED chips)
        Refreshrate: ~ 350Hz
        Flickers a little bit
    With 8 active bitplanes
        Brightness: bright (you don't the white LED chips)
        Refreshrate: ~ 200Hz
        No flickering seeable
    With 11 active bitplanes
        not possible

Results with EXPERIMENTAL_BITPLANES=11

    With 4 active bitplanes
        Brightness: bright (you don't the white LED chips)
        Refreshrate: ~ 100Hz
        Flickers a little bit
    With 8 active bitplanes
        Brightness: bright (you don't the white LED chips)
        Refreshrate: ~ 80Hz
        Flickers a little bit
    With 11 active bitplanes
        Brightness: low (but viewable) - you can see the white LED chips
        Refreshrate: ~ 75Hz
        Flickers a little bit

Conclusion:
The lower the maximum available bitplanes the darker the LEDs are. But the better the refresh rate will be.
So the user should experiment with this parameter to get the best value for his own needs.
For me for example I think the value 5 or 6 will be the best choice between number of colors, brightness of the LEDs (with 8 or 11 bitplanes my panel is too bright) and refreshrate

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -56,13 +56,6 @@ DEFINES+=-DEXPERIMENTAL_NEW_PINOUT
 # If the new pinout is used we should be able to set the maximum available
 # PWM bits. Lower available PMW bits will result in better refresh rate.
 # Little sideffect: the smaller the value the darker the LEDs will be.
-# 
-# Testing commandline:
-#    ./led-matrix -r 32 -P 1 -c 1 -p 4 -D 4
-#
-# Results with EXPERIMENTAL_BITPLANES 4
-# Results with EXPERIMENTAL_BITPLANES 8
-# Results with EXPERIMENTAL_BITPLANES 11
 DEFINES+=-DEXPERIMENTAL_BITPLANES=4
 
 INCDIR=../include
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -52,6 +52,19 @@ TARGET=librgbmatrix.a
 # while working on the new pinout.
 DEFINES+=-DEXPERIMENTAL_NEW_PINOUT
 
+# !!! Only tested with new pinout !!!
+# If the new pinout is used we should be able to set the maximum available
+# PWM bits. Lower available PMW bits will result in better refresh rate.
+# Little sideffect: the smaller the value the darker the LEDs will be.
+# 
+# Testing commandline:
+#    ./led-matrix -r 32 -P 1 -c 1 -p 4 -D 4
+#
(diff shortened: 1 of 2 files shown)


hzeller:
How is that different from the
  -p <pwm-bits>
option ?

hzeller:
I see, you are shifting the brightness down with that. I think that should be a commandline option.

Saij:
I think that too. I'm trying to implement it.
But there should be more experiments. I only saw a difference in brightness with the value 4 and 5 for this parameter. I think the problem here is that the PWMs with lower values are faster than clocking in the next row - so that there is more black time.
What about the base bitplane pwm time (currently 200ns) - maybe there is also a chance that changing this value will result in a brightness difference.

Currently I can't test it (building up a mobile Raspberry PI playground box with RT kernel ^^)

hzeller:
FYI: The head of that experimental branch already uses 100ns for the pwm time.

If you have some changes, consider doing them in the master branch as I am messing with the experimental branches quite a bit right now, and it might be messy to patch while work is in progress; but once I down-integrate things back to master, I can deal with that.

Saij:
Oh! Didn't see that with 100ns. Will try it out if it works better.
Maybe I'll fork the experimental-pwm branch so I'll get a stable version for my own ^^

Saij:
Will close that Pull Request (messed a little bit with my branches) cause will put that DEFINE to a command line parameter

PR #43, opened 2015-07-05
 Brightness Control

Saij:
Added functions to control brightness of LED Matrix

Currently SetBrightness() has to be called BEFORE you create a FrameCanvas and BEFORE you draw anything on this FrameCanvas (or the matrix).
Next step will be to make a dynamic brightness control without the need to redraw manually (maybe it will be better to store the RGB data additionally in the canvas and reset the bitplane data after a change on the brightness. So it also will be possible to copy a canvas for better front- and backbuffer handling)

### DIFF
--- a/demo-main.cc
+++ b/demo-main.cc
@@ -96,6 +96,7 @@ public:
       off_screen_canvas_ = matrix_->SwapOnVSync(off_screen_canvas_);
     }
   }
+
 private:
   RGBMatrix *const matrix_;
   FrameCanvas *off_screen_canvas_;
@@ -106,7 +107,7 @@ class BrightnessPulseGenerator : public ThreadedCanvasManipulator {
 public:
   BrightnessPulseGenerator(RGBMatrix *m) : ThreadedCanvasManipulator(m), matrix_(m) {}
   void Run() {
-    uint8_t max_brightness = matrix_->brightness();
+    const uint8_t max_brightness = matrix_->brightness();
     uint8_t count = 0;
     uint8_t c = 255;
 
@@ -128,6 +129,7 @@ public:
       usleep(20 * 1000);
     }
   }
+
 private:
   RGBMatrix *const matrix_;
 };
@@ -1057,7 +1059,7 @@ static int usage(const char *progname) {
           "\t                   terminal (e.g. cron).\n"
           "\t-t <seconds>     : Run for these number of seconds, then exit.\n"
           "\t                   (if neither -d nor -t are supplied, waits for <RETURN>)\n"
-          "\t-b <brightness>  : Sets brightness level. Default: 1. Range: 1..100\n");
+          "\t-b <brightness>  : Sets brightness level. Default: 100. Range: 1..100\n");
   fprintf(stderr, "Demos, choosen with -D\n");
   fprintf(stderr, "\t0  - some rotating square\n"
           "\t1  - forward scrolling an image (-m <scroll-ms>)\n"
--- a/demo-main.cc
+++ b/demo-main.cc
@@ -101,6 +101,37 @@ private:
   FrameCanvas *off_screen_canvas_;
 };
 
+// Simple generator that pulses through brightness on red, green, blue and white 
+class BrightnessPulseGenerator : public ThreadedCanvasManipulator {
+public:
+  BrightnessPulseGenerator(RGBMatrix *m) : ThreadedCanvasManipulator(m), matrix_(m) {}
+  void Run() {
+    uint8_t max_brightness = matrix_->brightness();
+    uint8_t count = 0;
+    uint8_t c = 255;
+
+    while (running()) {
+      if (matrix_->brightness() < 1) {
+        matrix_->SetBrightness(max_brightness);
+        count++;
+      } else {
+        matrix_->SetBrightness(matrix_->brightness() - 1);
+      }
+
+      switch (count % 4) {
+        case 0: matrix_->Fill(c, 0, 0); break;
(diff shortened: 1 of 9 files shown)


include/led-matrix.h
```diff
@@ -84,6 +84,9 @@ class RGBMatrix : public Canvas {
   void set_luminance_correct(bool on);
   bool luminance_correct() const;
 
+  void SetBrightness(uint8_t brightness);
```
hzeller: It might make sense to add the brightness setting as well to the FrameCanvas

lib/led-matrix.cc
```diff
@@ -159,7 +161,7 @@ bool RGBMatrix::SetPWMBits(uint8_t value) {
   }
   return success;
 }
-uint8_t RGBMatrix::pwmbits() { return active_->framebuffer()->pwmbits(); }
+uint8_t RGBMatrix::pwmbits() { return pwm_bits_; }
```
hzeller: Why did you change this ? (also for luminance correct).
The contract is, that these return the value for the currently active buffer.

demo-main.cc
```diff
@@ -101,6 +101,37 @@ class ColorPulseGenerator : public ThreadedCanvasManipulator {
...
+        case 0: matrix_->Fill(c, 0, 0); break;
+        case 1: matrix_->Fill(0, c, 0); break;
+        case 2: matrix_->Fill(0, 0, c); break;
+        case 3: matrix_->Fill(c, c, c); break;
+      }
+
+      usleep(20 * 1000);
+    }
+  }
+private:
```
hzeller: Can you add a newline before the "private:" ?

demo-main.cc
```diff
@@ -101,6 +101,37 @@ class ColorPulseGenerator : public ThreadedCanvasManipulator {
   FrameCanvas *off_screen_canvas_;
 };
 
+// Simple generator that pulses through brightness on red, green, blue and white 
+class BrightnessPulseGenerator : public ThreadedCanvasManipulator {
+public:
+  BrightnessPulseGenerator(RGBMatrix *m) : ThreadedCanvasManipulator(m), matrix_(m) {}
+  void Run() {
+    uint8_t max_brightness = matrix_->brightness();
```
hzeller: max_brightness can be 'const'. Also 'c'.

demo-main.cc
```diff
@@ -1016,17 +1047,17 @@ static int usage(const char *progname) {
...
+          "\t-L               : 'Large' display, composed out of 4 times 32x32\n"
+          "\t-p <pwm-bits>    : Bits used for PWM. Something between 1..11\n"
+          "\t-l               : Don't do luminance correction (CIE1931)\n"
+          "\t-D <demo-nr>     : Always needs to be set\n"
+          "\t-d               : run as daemon. Use this when starting in\n"
+          "\t                   /etc/init.d, but also when running without\n"
+          "\t                   terminal (e.g. cron).\n"
+          "\t-t <seconds>     : Run for these number of seconds, then exit.\n"
+          "\t                   (if neither -d nor -t are supplied, waits for <RETURN>)\n"
+          "\t-b <brightness>  : Sets brightness level. Default: 1. Range: 1..100\n");
```
hzeller: Default is 100, isn't it ?

python/samples/samplebase.py
```diff
@@ -10,6 +10,7 @@ def __init__(self, *args, **kwargs):
         self.add_argument("-c", "--chain", action = "store", help = "Daisy-chained boards. Default: 1.", default = 1, type = int)
         self.add_argument("-p", "--pwmbits", action = "store", help = "Bits used for PWM. Something between 1..11. Default: 11", default = 11, type = int)
         self.add_argument("-l", "--luminance", action = "store_true", help = "Don't do luminance correction (CIE1931)")
+        self.add_argument("-b", "--brightness", action = "store", help = "Sets brightness level. Default: 1. Range: 1..100", default = 100, type = int)
```
hzeller: Same here, default is 100

hzeller:
This looks mostly good, I only have some small comments , see above.

Also, instead of relative to experimental-pwm-bit, can you make this a pull request against the 'master' branch ?

demo-main.cc
Saij: OK. Will change it.
What do you mean by 'c'?

demo-main.cc
Saij: Will do.

demo-main.cc
Saij: Arg. Typo. Will change it (yes default should be 100% brightness)

include/led-matrix.h
Saij: The problem here is, that if you draw something and then change the brightness, only new drawings will get that new brightness. That's why I was not sure if implementing the function in FrameCanvas is such a good idea (also brightness should be a global setting to the device like luminance correction) and not apply to a single frame.

lib/led-matrix.cc
Saij: Because I think it should provide the current setting and not for the active frame.
The active frame could have not the same value than the global value set in the matrix.
Maybe it would be an idea to implement a getter function for the active framecanvas where you can get the information for the active frame.
But here it should return the value currently set to the matrix (normally the same as for the current frame).

python/samples/samplebase.py
Saij: Will change it.

Saij:
I can make the pull request against the master branch, but I can't test it on the master branch. I currently have a fixed setup for the PWM Pin Out.

Saij:
Closing this pull request to make a pull request against master branch

PR #44, opened 2015-07-06
Brightness Control

Saij:
Added functions to control brightness of LED Matrix

Currently SetBrightness() has to be called BEFORE you create a FrameCanvas and BEFORE you draw anything on this FrameCanvas (or the matrix).
Next step will be to make a dynamic brightness control without the need to redraw manually (maybe it will be better to store the RGB data additionally in the canvas and reset the bitplane data after a change on the brightness. So it also will be possible to copy a canvas for better front- and backbuffer handling)

(New pull request for master branch. Old request can be found under following URL https://github.com/hzeller/rpi-rgb-led-matrix/pull/43)

### DIFF
--- a/demo-main.cc
+++ b/demo-main.cc
@@ -108,8 +108,8 @@ public:
   BrightnessPulseGenerator(RGBMatrix *m) : ThreadedCanvasManipulator(m), matrix_(m) {}
   void Run() {
     const uint8_t max_brightness = matrix_->brightness();
+    const uint8_t c = 255;
     uint8_t count = 0;
-    uint8_t c = 255;
 
     while (running()) {
       if (matrix_->brightness() < 1) {
@@ -1059,7 +1059,7 @@ static int usage(const char *progname) {
           "\t                   terminal (e.g. cron).\n"
           "\t-t <seconds>     : Run for these number of seconds, then exit.\n"
           "\t                   (if neither -d nor -t are supplied, waits for <RETURN>)\n"
-          "\t-b <brightness>  : Sets brightness level. Default: 100. Range: 1..100\n");
+          "\t-b <brightness>  : Sets brightness percent. Default: 100.\n");
   fprintf(stderr, "Demos, choosen with -D\n");
   fprintf(stderr, "\t0  - some rotating square\n"
           "\t1  - forward scrolling an image (-m <scroll-ms>)\n"
@@ -1181,7 +1181,7 @@ int main(int argc, char *argv[]) {
   }
 
   if (brightness < 1 || brightness > 100) {
-    fprintf(stderr, "Brightness is oudside usable range.\n");
+    fprintf(stderr, "Brightness is outside usable range.\n");
     return 1;
   }
 
--- a/demo-main.cc
+++ b/demo-main.cc
@@ -96,6 +96,7 @@ public:
       off_screen_canvas_ = matrix_->SwapOnVSync(off_screen_canvas_);
     }
   }
+  
 private:
   RGBMatrix *const matrix_;
   FrameCanvas *off_screen_canvas_;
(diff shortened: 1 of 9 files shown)


demo-main.cc
```diff
@@ -1141,6 +1180,11 @@ int main(int argc, char *argv[]) {
     return 1;
   }
 
+  if (brightness < 1 || brightness > 100) {
+    fprintf(stderr, "Brightness is oudside usable range.\n");
```
hzeller: s/oudside/outsite/

demo-main.cc
```diff
@@ -1016,17 +1049,17 @@ static int usage(const char *progname) {
...
+          "\t-L               : 'Large' display, composed out of 4 times 32x32\n"
+          "\t-p <pwm-bits>    : Bits used for PWM. Something between 1..11\n"
+          "\t-l               : Don't do luminance correction (CIE1931)\n"
+          "\t-D <demo-nr>     : Always needs to be set\n"
+          "\t-d               : run as daemon. Use this when starting in\n"
+          "\t                   /etc/init.d, but also when running without\n"
+          "\t                   terminal (e.g. cron).\n"
+          "\t-t <seconds>     : Run for these number of seconds, then exit.\n"
+          "\t                   (if neither -d nor -t are supplied, waits for <RETURN>)\n"
+          "\t-b <brightness>  : Sets brightness level. Default: 100. Range: 1..100\n");
```
hzeller: Maybe we should call it 'Sets brightness percent'. Then people can easily relate that this is a 1..100 range.

lib/framebuffer.cc
```diff
@@ -148,16 +148,18 @@ inline Framebuffer::IoBits *Framebuffer::ValueAt(int double_row,
 }
 
 // Do CIE1931 luminance correction and scale to output bitplanes
-static uint16_t luminance_cie1931(uint8_t c) {
+static uint16_t luminance_cie1931(uint8_t c, uint8_t brigtness) {
```
hzeller: s/brigtness/brightness/

hzeller:
Only some comments (mostly typos), then it is ready to merge.

I think I wouldn't worry about dynamic brightness change at this point; I think it is sufficiently ok to tell users that brightness only affects newly drawn things (a backing buffer that does a re-draw could easily be implemented with a stacked Canvas on top that keeps track of the pixels).

demo-main.cc
```diff
@@ -96,11 +96,44 @@ class ColorPulseGenerator : public ThreadedCanvasManipulator {
...
 };
 
+// Simple generator that pulses through brightness on red, green, blue and white 
+class BrightnessPulseGenerator : public ThreadedCanvasManipulator {
+public:
+  BrightnessPulseGenerator(RGBMatrix *m) : ThreadedCanvasManipulator(m), matrix_(m) {}
+  void Run() {
+    const uint8_t max_brightness = matrix_->brightness();
+    uint8_t count = 0;
+    uint8_t c = 255;
```
hzeller: With 'c' I meant, that this variable can be const as well.

Saij:
The thing with the stacked Canvas on top is planned next for me ^^
Already have an idea how to implement an "extended" main class that keeps track of pixels and backbuffer handling (need such class in C++ cause in Python it would be way too slow)

Saij:
Fixed the typos and added the missing "const" declaration.

hzeller:
Uhm, did you ever test this ? The luminance array was accessed way out of bounds ...

Fixed in
https://github.com/hzeller/rpi-rgb-led-matrix/commit/c88c9653305ec9e99acdca74782b220616134188

Saij:
Yeah. Tested it. But got problems merging the changes in the PWM branch into the master branch. Maybe there I got the Index Out Of Bounds Problem

Saij:
Don't know what went wrong. Checked my repo and there was this error too. But the demo (-D 11) worked fine. Also the other demos worked fine if I used the -b parameter.
I shove the blame on the heat ^^

PR #46, opened 2015-07-08
Added Coordinate Mapper Callback Support

Saij:
I added support for coordination mapping without the need to create a wrapper class and create an object of that wrapper class each time you create a new FrameCanvas object. I think they are easier to handle and easier to understand. The old method still works and can be used.
It also allows for multiple coordination mapping.

As example there are 3 coordination mapping callback:

1. Rotate Coordination Mapper
    Rotates the whole matrix 90°, 180° or 270°

2. Large Canvas Support
    Replaces wrapper class LargeSquare64x64Canvas to support square 64x64 canvas with 4 32x32 matrices (or 2 64x32) ordered in C shape (and not parallel)

3. Stretch Coordination Mapper
    Just a simple mapper that multiplies the x coordinate with 2. (Needed this to test the multiple coordination mapping cause I only have one 32x32 matrix)

The callbacks are stored in std::vector lists and are processed First Added - First Called.
So you can first add the rotate mapper and then you could add the LargeCanvas mapper to rotate a 64x64 matrix. If you reverse the order, the result will be very funny (first the LargeCanvas mapper maps the 64x64 coordinates to a 128x32 system and then this 128x32 system will be rotatet).

As statet couldn't test the LargeCanvas mapper but I took your code out of the class. So I think it will be running. If it works correctly with the rotate mapper I can't say for sure.

Also I've done some small works like adding the new parameters to README.md and so on ^^

### DIFF
--- a/include/canvas.h
+++ b/include/canvas.h
@@ -46,5 +46,34 @@ public:
   virtual void Fill(uint8_t red, uint8_t green, uint8_t blue) = 0;
 };
 
+// A canvas transformer is an object that, given a Canvas, returns a
+// canvas that applies transformations before writing to the original
+// Canvas.
+//
+// To simplify assumptions for implementations and users of implementations,
+// the following conditions and constraints apply:
+//
+//   1) The CanvasTransformer _never_ takes ownership of the delegatee.
+//   2) The ownership of the returned Canvas is _not_ passed to the caller.
+//   3) The returned Canvas can only be assumed to be valid for the lifetime
+//      of the CanvasTranformer and the lifetime of the original canvas.
+//   4) The returned canvas is only valid up to the next call of Transform().
+//
+//   * Point 2)-4) imply that the CanvasTransformer can hand out the same
+//     canvas object every time, just configured to write to the new
+//     delegatee.
+//   * Point 4) implies that one instance of CanvasTransformer cannot be used in
+//     parallel in multiple threads calling Transform().
+//   * The constraints also imply that it is valid for a CanvasTransformer to
+//     return the passed in canvas itself.
+class CanvasTransformer {
+public:
+  virtual ~CanvasTransformer() {}
+
+  // Return a Canvas* that applies transformations before delegating to
+  // the output canvas.
+  virtual Canvas *Transform(Canvas *output) = 0;
+};
+
 }  // namespace rgb_matrix
 #endif  // RPI_CANVAS_H


hzeller:
I am not convinced that it is a good idea to have a coordination mapper function as part of the matrix or FrameCanvas. I think in particular it will be more troublesome than useful.

It mixes concerns in a way that makes the base RGBMatrix/FrameCanvas canvases more complex while it only addresses a particular aspect of what you might want to do when mapping your user-canvas to some physical canvas (there is color, transparency, scaling etc ..).

Also a single function doesn't cover the whole picture (pun intended): not only coordinates but also dimensions are typically mapped. Consider the LargeSquare64x64Canvas. It maps a single chain with four panels (so of size 128x32) to a square panel with 64x64. However, I can't ask the canvas what size it has after the mapping, as it is still 128x32 instead of the new logical size.

I think stuff gets even more complicated for a user because they have to modify in/out parameters. I think it would be better if the function would return an x/y tuple of some kind or have two input parameters and two output parameters; modifying parameters results in less readable code.

There are other smaller practical problems: in your example, you have the angle, and pivot_x/y, which now need to be global variables as you can't do currying with simple means (it could be improved if we use a std::function in C++11 environments; because an inline-mapping function might be quite neat). This is the kind of configuration information that is usually better encapsulated in an object.

The Scaling example is an example where things fall short with simple mapping: if you scale something up or down, you need to modify the adjacent pixels as well. In your stretch example, every other column would be empty, but it should have the average between left and right pixel column.
That only works if you wrap a whole Canvas around that can do the operation.

I don't particular like the stacking of vector<Callback> as opposed to stacking of Canvas objects delegating to each other which is easier to reason about.

If you want to implement something simple that allows you to pass a callback, you should have a separate Canvas implementation that does all the boilerplate, and you pass it the callback in the constructor:
  class CoordinateMapperCanvas : public Canvas {
   public:
      CoordinateMapperCanvas(Canvas *delegatee, CoordinateMapperCallback *cb, int new_x, int new_y);
     void SetDelegatee(Canvas *delegatee);  // possible to change deleatee on demand
  }

To have things easy chainable and readable, you can also add some convenience function:
   Canvas *CoordinateMapperCanvas::MapCanvas *delegatee) {
       SetDelegatee(delegatee);
       return this;
  }

Then you can do things like this:
CoordinateMapperCanvas mapper(NULL, callback, 64, 64);
FunctionThatTakesACanvasPointer(mapper.Map(current_frame_canvas));
... that is helpful in the common case with multi-buffering, but you still only need to allocate one mapper canvas.

This would be more readable.

hzeller:
In https://github.com/hzeller/rpi-rgb-led-matrix/commit/83b8bf4bc26043a8b62fbd002ed5174fbb8d7a5f I'ave now added a CanvasTransformer interface that provides a method 
  Canvas *Transform(Canvas *output);

(which is similar to the suggested Map() in my previous comment).

So now, you could implement a callback mapper the following way.
````
  CallbackCoordinateMapper : public CanvasTransformer {
      CallbackCoordinateMapper(CoordinateMapperCallback *callback, int new_x, int new_y);
      virtual Canvas *Transform(Canvas *output) { /* set output to map_canvas_*/ return map_canvas_; }
  private:
     class MappingCanvasImplementation;  // only needs to be visible in compilation unit.
     MappingCanvasImplementation *map_canvas_;
  }
 ````

and use it in context such as
```
   CallbackCoordinateMapper mapper(callback, 64, 64);
   // ...
   FunctionThatTakesACanvasPointer(mapper.Transform(current_frame_canvas));
```

**Comment by Saij** [comment]
You are right regarding things like width() and height(). Didn't took those function into account while developing.

I don't see the problem with the prototype. But this is the smallest thing to change (as it is only a prototype).

Regarding the stretch mapper: this was intended to leave one row always black. It was just a small test ot check if the stack is working correctly as I couldn't test rotate with large canvas.

My big problem with stacking the Canvas objects is simple: the Canvas interface doesn't implement functions like CreateFrameCanvas and SwapOnVSync. This is correct but in the case of stacking not so good to use.
For example the demo-main.cc: every demo in there that uses double buffering needs an instance of RGBMatrix to be passed to create the needed buffers. Now to handle things like a large canvas or rotate it would also need a list of class names for the canvas stack to wrap each new canvas in the required objects.
In my opinion it would complicate the code more than a small callback (but a callback also isn't the right approach as you stated).

Even with your approach of a CanvasTransform class you would need a second parameter. If, for example, you want a stretch, a rotate and a large canvas transformation, you would need a LargeCanvasTransform which is encapsulated in a RotateCanvasTransform which is encapsulated in a StretchTransform. If you then pass the StretchTransform into the demo classes (as the stretch should be the first thing in the transformation stack), how to set the delegatee Canvas (created by CreateFrameCanvas) to the LargeCanvasTransform (because the created canvas is the result and thus need to be at the bottom of the stack)?
I think it is better to set transformation, and other things that need to be done before the actual drawing takes place, between the canvas interface and the framebuffer object. So you only need to set those globally (in case of the large canvas it simply is global to the matrix as the matrix don't change between framechanges).

One idea would maybe injector objects. For example you have following class:
```c++
class SampleCanvasTransformInjector {
public:
    void SetPixel(int x, int y, Canvas* destinationCanvas);
    void width(Canvas* destinationCanvas);
    void height(Canvas* destinationCanvas);
}
```

Framebuffer then only needs to implement the Canvas interface to support the required functions. Then you just inject an object of SampleCanvasTransformInjector like the callback into the matrix and instead of calling the framebuffer methods from the FrameCanvas class you call the corresponding function of the SampleCanvasTransformInjector passing in the framebuffer (which now is a subtype of Canvas).

(Hope you understood what I mean ^^)

hzeller:
The demo classes don't need a list of class names, only one CanvasTransform object that they can apply to the frame canvases.

If you need multiple transformations done, you still can have a CanvasTranform implementation that gets a list of CanvasTransform objects that are then called in sequence (well, there the need for a vector re-emerges, but it is in a separate class, so somewhat ok).

With that stacking, you can solve as well the problem where you think you need a second parameter; you actually don't; say the LinkedTransformation is essentially doing the following:
    LinkedTransformation::Transform(Canvas *c) { return stretch->Transform(rotate->Transform(large_canvas->Transform(c))); }

The Canvas *c you pass in can be the FrameCanvas, which is at the very end of the 'food chain'.

As you state, it might be worthwhile considering adding a CanvasTransformer to the RGBMatrix that then globally does the mapping to all FrameCanvases it hands out (though it would require to narrow the thread-safety constraints on the CanvasTransformer).

Looking forward to your next iteration :)

Saij:
Just to be sure I understood your concept I wrote a little (not working) example.
Could you please tell me if it is in your concept or did I mess up again?

```c++
class LargeSquare64x64Canvas : public Canvas {
public:
  LargeSquare64x64Canvas() : delegatee_(NULL) {}
  virtual ~LargeSquare64x64Canvas() { delete delegatee_; }

  void SetDelegatee(Canvas* delegatee) {
    assert(delegatee != NULL);
    // Our assumptions of the underlying geometry:
    assert(delegatee->height() == 32);
    assert(delegatee->width() == 128);

    delegatee_ = delegatee;
  }

  virtual void Clear() { assert(delegatee_ != NULL); delegatee_->Clear(); }
  virtual void Fill(uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);
    delegatee_->Fill(red, green, blue);
  }
  virtual int width() const { return 64; }
  virtual int height() const { return 64; }
  virtual void SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);

    if (x < 0 || x >= width() || y < 0 || y >= height()) return;
    // We have up to column 64 one direction, then folding around. Lets map
    if (y > 31) {
      x = 127 - x;
      y = 63 - y;
    }
    delegatee_->SetPixel(x, y, red, green, blue);
  }

private:
  Canvas *delegatee_;
};

class LargeSquare64x64CanvasTransform : public CanvasTransformer {
public:
  LargeSquare64x64CanvasTransform() {
    canvas_ = new LargeSquare64x64Canvas();
  }

  virtual Canvas *Transform(Canvas *output) {
    canvas_->SetDelegatee(output);
    return canvas_;
  }

private:
  LargeSquare64x64Canvas canvas_;
};

class StretchCanvas : public Canvas {
public:
  // This class takes over ownership of the delegatee.
  StretchCanvas() : delegatee_(NULL) {}
  virtual ~StretchCanvas() { delete delegatee_; }

  void SetDelegatee(Canvas* delegatee) {
    assert(delegatee != NULL);
    delegatee_ = delegatee;
  }

  virtual void Clear() { assert(delegatee_ != NULL); delegatee_->Clear(); }
  virtual void Fill(uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);
    delegatee_->Fill(red, green, blue);
  }
  virtual int width() const { assert(delegatee_ != NULL); return delegatee_->width(); }
  virtual int height() const { assert(delegatee_ != NULL); return delegatee_->height(); }
  virtual void SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);
    delegatee_->SetPixel(x * 2, y, red, green, blue);
  }

private:
  Canvas *delegatee_;
};

class StretchCanvasTransform : public CanvasTransformer {
public:
  StretchCanvasTransform() {
    canvas_ = new StretchCanvas();
  }

  virtual Canvas *Transform(Canvas *output) {
    canvas_->SetDelegatee(output);
    return canvas_;
  }

private:
  StretchCanvas canvas_;
};

class RotateCanvas : public Canvas {
public:
  // This class takes over ownership of the delegatee.
  RotateCanvas(int angle) : delegatee_(NULL), angle_(angle) {}
  virtual ~RotateCanvas() { delete delegatee_; }

  void SetDelegatee(Canvas* delegatee) {
    assert(delegatee != NULL);

    pivot_x_ = (delegatee->width() - 1) / 2;
    pivot_y_ = (delegatee->height() - 1) / 2;
    delegatee_ = delegatee;
  }

  virtual void Clear() { assert(delegatee_ != NULL); delegatee_->Clear(); }
  virtual void Fill(uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);
    delegatee_->Fill(red, green, blue);
  }
  virtual int width() const { assert(delegatee_ != NULL); return delegatee_->width(); }
  virtual int height() const { assert(delegatee_ != NULL); return delegatee_->height(); }
  virtual void SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue) {
    assert(delegatee_ != NULL);

    const float s = (angle_ == 90 ? 1 : (angle_ == 270 ? -1 : 0));
    const float c = (angle_ == 180 ? -1 : (angle_ == 0 ? 1 : 0));

    // Offset needed cause of little precision errors on 180° and 270°
    const int offset_x = (angle_ == 90 || angle_ == 180 ? 1 : 0);
    const int offset_y = (angle_ == 180 || angle_ == 270 ? 1 : 0);

    // translate point to origin
    x -= pivot_x_;
    y -= pivot_y_;

    float rot_x = x * c - y * s;
    float rot_y = x * s + y * c;

    // translate back
    x = rot_x + pivot_x_ + offset_x;
    y = rot_y + pivot_y_ + offset_y;

    delegatee_->SetPixel(x, y, red, green, blue);
  }

private:
  Canvas *delegatee_;

  int angle_;
  float pivot_x_;
  float pivot_y_;
};

class RotateCanvasTransform : public CanvasTransformer {
public:
  RotateCanvasTransform(int angle) {
    canvas_ = new RotateCanvas(angle);
  }

  virtual Canvas *Transform(Canvas *output) {
    canvas_->SetDelegatee(output);
    return canvas_;
  }

private:
  RotateCanvas canvas_;
};

class DynamicLinkedTransformation : public CanvasTransformer {
public:
  virtual Canvas *Transform(Canvas *output) {
    for (size_t i = transformer_list_.size() - 1; i >= 0; i--) {
      output = transformer_list_[i]->Transform(output);
    }

    return output;
  }

  void AddTransformer(CanvasTransformer *transformer) {
    assert(transformer != NULL);
    transformer_list_.push_back(transformer);
  }

private:
  std::vector<CanvasTransformer> transformer_list_;
};

int main(int argc, char *argv[]) {
  // Initialize GPIO pins. This might fail when we don't have permissions.
  GPIO io;
  if (!io.Init())
    return 1;

  RGBMatrix *matrix = new RGBMatrix(&io, 32, 1, 1);

  RotateCanvasTransform *rotator = new RotateCanvasTransform(90 /* angle in degree */);
  StretchCanvasTransform *stretcher = new StretchCanvasTransform();
  LargeSquare64x64CanvasTransform *large = new LargeSquare64x64CanvasTransform();

  DynamicLinkedTransformation *transformer = new DynamicLinkedTransformation();
  
  // First add the LargeSquare64x64CanvasTransform cause it should be the last transformation done
  transformer->AddTransformer(large);

  // Then we add the stretcher
  transformer->AddTransformer(stretcher);

  // Last but not least we add the rotator
  transformer->AddTransformer(rotator);

  // Set Pixel directly to the matrix
  transformer->Transform(matrix)->SetPixel(2, 2, 255, 255, 255);

  // Now set a pixel on a backbuffer
  FrameCanvas* offbuffer = matrix->CreateFrameCanvas();
  transformer->Transform(offbuffer)->SetPixel(4, 4, 255, 255, 255);

  // This shouldn't work
  Canvas* buffer1 = transformer->Transform(matrix->CreateFrameCanvas());
  Canvas* buffer2 = transformer->Transform(matrix->CreateFrameCanvas());
  // Now buffer1 and buffer2 should point to the same object in memory
  // Don't think that this is so good
}
```

hzeller:
Yes essentially.

The
```
   transformer->Transform(matrix)->SetPixel(2, 2, 255, 255, 255);
```
situations work great with this, but keeping buffers around like the following of course not:
```
 Canvas* buffer1 = transformer->Transform(matrix->CreateFrameCanvas());
 Canvas* buffer2 = transformer->Transform(matrix->CreateFrameCanvas());
```

However, I would argue whenever you generate some graphics, you only do one canvas at a time, as things typically looks like this:
```
  for (;;) {
     transformer->Transform(current_frame)->SetPixel(...);
     current_frame = matrix->SwapOnVSync(current_frame);
  }
```

Generally I like that because you only use the object for what you need them for and still have a clean canvas interface that you can hand out somewhere, and don't create too many objects where not needed.

To cover the case that you want to keep multiple objects around, maybe the transformer class could also have a CreateTransform() or something that is a factory for a new Canvas. This would create a new object, but that then would be independent.

Of course, for simple transformations that don't require a lot of state to be prepared typically often as well cheap to just create a delegating canvas that does not take ownership on the stack.

```
   for (;;) {
      ConvertingCanvas canvas(current_frame);
      canvas.SetPixel(...);
      current_frame = matrix->SwapOnVSync(current_frame);
  }
```
... however, it is not good for 'passing in' some transformation.

Saij:
Ah now I've got the point.

Maybe also a good idea is to implement some standard CanvasTransform classes like the LargeCanvas in C style (with optional size arguments as there are different sizes out there like the 32x64 matrizes), the RotatingCanvas (where the angle should be a static member so that you can change the rotation without a new initalization) and the LinkedTransformCanvas (which only contains a list of CanvasTransformer like my example above).

If you are OK with that idea I will implement them. I hope that I can implement the same things into python without big effort.

Another question: why you don't like std::vector that much?

hzeller:
Regarding LargeCanvas, I already have a version cooking that I need to test and push that takes a text-description of the panel layout and provides the mapping. That should make it simple to set up things easily. So don't waste too much time converting that guy.

I like the idea of these standard transformers. For the rotating one, we might want to limit to multiple of 90 degrees (I think you already did that in the callback version). Otherwise we need to take aliasing into account to create a more pleasing output for weird angles; but that requires to possibly set multiple values, which would be problematic without backing buffer ...

The LinkedTransformCanvas can have a default constructor and add methods, and another one that already gets a std::vector<CanvasTransformer*>.
(if you are in C++11, this then could be made very readable with in-line initialization
   LinkedTransformer transformer({ new FooTransformer(), new BarTransformer() });
)

I do like std::vector<>, but sometimes things are bit more nicely expressed in a functional style, in particular these kind of transformations (so calling a function that calls a function that calls a function). However, for the LinkedTransformer, a vector is a fine choice.

Saij:
Then I'll implement these standard transformers except for the large canvas transformer.

hzeller:
SGTM

Saij:
Please? Don't know what it means ^^

hzeller:
SGTM = "sounds good to me" :)

PR #48, opened 2015-07-10
Implemented standard transformers

Saij:
Like in the other pull request discussed I implemented some standard transformers (RotateCanvasTransformer, LinkedTransformerCanvasTransformer and - even if you said that I don't have to - LargeSquare64x64CanvasTransformer).

Also I added an interface to RGBMatrix class to set a default transformer that the drawing functions of the RGBMatrix will use (so you don't have to create an extra canvas just for doing transformations).

Adapted the demos to use the new transformer interface instead directly writing to a canvas (but only if they use double buffering - else the RGBMatrix object with its default transformer will handle it).

Didn't implement a CreateTransformer() function cause I couldn't think of a usecase (and also got no idea how to access the angle property of the RotateCanvas in a good way).


lib/led-matrix.cc
```diff
@@ -183,16 +184,45 @@ uint8_t RGBMatrix::brightness() {
   return brightness_;
 }
 
+void RGBMatrix::SetTransformer(CanvasTransformer *transformer) {
+  transformer_ = transformer;
```
hzeller: I think I'd just keep a dummy transformer around that just returns the canvas itself when someone sets the transformer to NULL.

Otherwise you have to always check for transformer_ != NULL everywhere you want to use it, which is error prone and harder to read.

include/transformer.h
```diff
@@ -0,0 +1,118 @@
...
+
+private:
+  RotateCanvas *canvas_;
+  int angle_;
+};
+
+// Transformer for linked transformer objects
+// First transformer added will be considered last
+// (so it would the transformer that gets the original Canvas object)
+typedef std::vector<CanvasTransformer*> CanvasTransformerList;
```
hzeller: I'd put this typedef inside the LinkedTransformer thing itself. And maybe call it shorter, like 'List'.
Then the full name would be LinkedTransformerCanvasTransformer::List.

include/transformer.h
```diff
@@ -0,0 +1,118 @@
...
+
+#include <vector>
+#include <cstddef>
+
+#include "canvas.h"
+
+namespace rgb_matrix {
+
+// Transformer canvas to rotate the input canvas in 90° steps
+class RotateCanvas : public Canvas {
```
hzeller: I'd not put this in the header file but either in an anonymous namespace in the compilation unit or as a RotateCanvasTransformer::RotateCanvas private class.

include/transformer.h
```diff
@@ -0,0 +1,118 @@
...
+
+private:
+  Canvas *delegatee_;
+  int angle_;
+  float pivot_x_;
+  float pivot_y_;
+};
+
+// Transformer for RotateCanvas
+class RotateCanvasTransformer : public CanvasTransformer {
```
hzeller: Maybe these names should be shortened without the 'CanvasTransformer' as suffix. So this one should probably just be 'RotateTransformer', the others 'LinkedTransformer' below and 'LargeSquare64x64Transformer'.

include/transformer.h
```diff
@@ -0,0 +1,118 @@
...
+public:
+  RotateCanvasTransformer(int angle);
+  virtual ~RotateCanvasTransformer() { delete canvas_; }
```
hzeller: Careful, you shouldn't take ownership of the delegatee.

include/transformer.h
```diff
@@ -0,0 +1,118 @@
...
+private:
+  CanvasTransformerList transformer_list_;
+};
+
+class LargeSquare64x64Canvas : public Canvas {
+public:
+  LargeSquare64x64Canvas() : delegatee_(NULL) {}
+  virtual ~LargeSquare64x64Canvas() { delete delegatee_; }
+
+  void SetDelegatee(Canvas* delegatee);
```
hzeller: Same here, no ownership taking.

lib/transformer.cc
```diff
@@ -0,0 +1,176 @@
...
+}
+
+void RotateCanvas::SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue) {
+  assert(delegatee_ != NULL);
+
+  const float s = (angle_ == 90 ? 1 : (angle_ == 270 ? -1 : 0));
+  const float c = (angle_ == 180 ? -1 : (angle_ == 0 ? 1 : 0));
+
+  // Offset needed cause of little precision errors on 180° and 270°
+  const int offset_x = (angle_ == 90 || angle_ == 180 ? 1 : 0);
```
hzeller: mmh, is this really needed ? I think the problem might be that you are not rounding properly when you assign the coordinates back to integers (because they fill be implicitly floor()-ed. So you should apply roundf() when translating back; then I don't think you need this offset fudging.

lib/transformer.cc
```diff
@@ -0,0 +1,176 @@
...
+
+int RotateCanvasTransformer::angle() {
+  return angle_;
+}
+
+void LinkedTransformerCanvasTransformer::AddTransformer(CanvasTransformer *transformer) {
+  transformer_list_.push_back(transformer);
+}
+
+void LinkedTransformerCanvasTransformer::AddTransformer(CanvasTransformerList transformer_list) {
```
hzeller: I'd name the parameter suff sufficiently differently that it is hard to mix up transformer_list and transformer_list_.

(just 'list' is probably a sufficiently good name).

lib/transformer.cc
```diff
@@ -0,0 +1,176 @@
...
+
+#include <assert.h>
+
+#include "transformer.h"
+
+namespace rgb_matrix {
+
+RotateCanvas::RotateCanvas(int angle) : delegatee_(NULL), angle_(angle) {
+  // Only allow angles in 90° steps
+  assert(angle % 90 == 0);
```
hzeller: Does the output look really that bad with other angles, that we don't want to allow it ? We could allow all angles, but 'recommend' only using muliple of 90.

lib/transformer.cc
```diff
@@ -0,0 +1,176 @@
...
+}
+
+void RotateCanvas::Fill(uint8_t red, uint8_t green, uint8_t blue) {
+  assert(delegatee_ != NULL);
+  delegatee_->Fill(red, green, blue);
+}
+
+void RotateCanvas::SetAngle(int angle) {
+  assert(angle % 90 == 0);
+  angle_ = angle;
```
hzeller: As a little optimization suggestion, I'd already store the sin() and cos() of the angle (in particular if we are going to allow arbitrary angles). That also simplifies the code in the critical path.

lib/transformer.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  // translate back
+  x = rot_x + pivot_x_ + offset_x;
+  y = rot_y + pivot_y_ + offset_y;
+
+  delegatee_->SetPixel(x, y, red, green, blue);
+}
+
+int RotateCanvas::width() const { 
+  assert(delegatee_ != NULL);
```
hzeller: instead of testing for NULL on every function call, your transformer should do the right thing. Makes the code more readable.

include/led-matrix.h
```diff
@@ -103,6 +104,9 @@ class RGBMatrix : public Canvas {
   // animation.
   FrameCanvas *SwapOnVSync(FrameCanvas *other);
 
+  void SetTransformer(CanvasTransformer *transformer);
+  CanvasTransformer *transformer();
```
hzeller: You could make that an inline implementation here.

Also we should make sure that whatever we return here is never NULL to not put the burden of NULL checking to the enduser. We should just set transformer_ to an IdentityTransformer when SetTransformer() passes in a NULL.

demo-main.cc
```diff
@@ -328,7 +288,7 @@ class ImageScroller : public ThreadedCanvasManipulator {
         for (int y = 0; y < screen_height; ++y) {
           const Pixel &p = current_image_.getPixel(
                      (horizontal_position_ + x) % current_image_.width, y);
-          offscreen_->SetPixel(x, y, p.red, p.green, p.blue);
+          matrix_->transformer()->Transform(offscreen_)->SetPixel(x, y, p.red, p.green, p.blue);
```
hzeller: the screen_height and width you are iterating through here ... they should be queried as well through the transformer, no ?

hzeller:
I did a first round of review with a few comments.

lib/led-matrix.cc
Saij: Will add the linked transformer as standard transformer. If there is no other transformer inside the linked transformer then the input canvas will be returned.

include/transformer.h
Saij: So the RotateCanvas shouldn't be usable without the transformer?

include/transformer.h
Saij: Ups sorry.
Shouldn't only copy things but also read ^^

lib/transformer.cc
Saij: I've tried it many times. With round(), floor() and ceil(). But it was always one pixel off. So I defined the offset.

lib/transformer.cc
Saij: If oyu think so I'll remove the assert. But I think its not practicable to allow other angles.

lib/transformer.cc
Saij: So delete the assertion and with a private RotateCanvas we should always think there is a canvas there. So only test against NULL in the Transform() function.

demo-main.cc
Saij: Ups forgot that (was late at night ^^)

lib/led-matrix.cc
Saij: Correction: implented a NullTransformer private to RGBMatrix class

include/transformer.h
Saij: Done

include/transformer.h
Saij: Implemented

include/transformer.h
Saij: Done

include/transformer.h
Saij: Fixed

include/transformer.h
Saij: Fixed

lib/transformer.cc
Saij: Done

lib/transformer.cc
Saij: Removed it - but didn't test the result with other angles

lib/transformer.cc
Saij: Precaluclated those values and stored them in class variables

lib/transformer.cc
Saij: Done

include/led-matrix.h
Saij: Done

demo-main.cc
Saij: Fixed

Saij:
Fixed things (or added a comment with my opinion why not to change it).

Also I think that the pwm branch is stable enough to merge it into the master branch. It is fully backwards compatible and if the user don't like the higher cpu usage he can go back to the old way by simply change a define value in the makefile.
I for myself are fine with the current state it is in. Even made a pinout that only works with the pwm branch (that's why I have a little trouble on the pull requests ^^ developing in the pwm branch and then merging only those changes into the master branch ^^)

hzeller:
Thanks I'll have a look as soon as possible (which might be tonight or tomorrow night).

There are a few issues with the PWM branch why I don't want to merge it with main yet
  - it creates yet another layout. To be backward compatible I need to support all of them with compile flags, but I want to keep that to a minimum. I want to see if I can avoid the explicit need for a particular pin in the DMA mode. Even if not, there is one pin that I still consider changing to be compatible with Revision A.
  - I need to update documentation regarding the pinout.
  - the PWM branch uses _more_ CPU than the master branch. On Raspberry Pi2 this is still ok-ish, on Raspberry Pi1 this is borderline unusable. Once DMA is working properly, that would be solved.

Long story short: I won't update the master branch with that code until the DMA code is ready. I hope to get it done in free time till the Chaos Communication Camp.

hzeller:
Sorry, I didn't get a chance yet to look at your pull request more closely, but I'll merge it soon.

My weekend was still busy with rgb matrix stuff:

I spent some time testing and documenting if DMA would be an option and sadly it isn't, because it won't be fast enough ( https://github.com/hzeller/rpi-gpio-dma-demo ).

So now I merged the PWM branch to master and updated the documentation (have a look at my new wiring diagram, that should be much easier to understand with less errors). It still has some problems (most notably, it is using more CPU and generates a lower frame-rate due to some internal reasons of how the PWM hardware is set-up. I need to read the datasheet more to come to a good solution that I hope to be able to fix soon).

Saij:
Please don't merge it now. I just crashed the python module while merging your master to my own master -.-

Saij:
Fixed it.
Only forgot to add the core.cpp. Now all python scripts will work again

Saij:
And now I wonder why this branch is marked "up-to-date"... these two branches (hzeller/rpi-rgb-led-matrix/master and vonaffenfels/rpi-rgb-led-matrix/master) are different...

hzeller:
Cool. I think the 'up-to-date' just means that your branch has a clear predecessor, which makes it easy to merge.
 
To help me write a commit message, I suppose your branch can be mostly considered having two themes:
  - create standard transformers
  - Update python directory structure
right ?

Saij:
Eeyup thats correct.
Sorry that I've made such a mess. Didn't know that committing to the branch also updates the pull request.

Saij:
Also added the graphic functions in the last commit.

hzeller:
Ok. Will do the merge tonight so that this is not pending too long.

include/transformer.h
```diff
@@ -0,0 +1,117 @@
...
+
+class LargeSquare64x64Transformer : public CanvasTransformer {
+public:
+  LargeSquare64x64Transformer();
+  virtual ~LargeSquare64x64Transformer() { delete canvas_; };
+
+  virtual Canvas *Transform(Canvas *output);
+
+private:
+  class LargeSquare64x64Canvas : public Canvas {
```
hzeller: I'd not put the whole class declaration in here, you can just to that in the compilation unit. That way, you keep the header file small.

So
  private:
     class LargeSquare64x64Canvas;
     LargeSquare64x64Canvas *canvas_;

here. And in the compilation unit:
 class LargeSquare64x64Transformer::LargeSquare64x64Canvas {
       // implementation.
 };

That way, the header file is relatively small. For instance RotateCanvas uses quite some space with all its instance variables.

In general, I'd probably just call the internal canvas of all the transformers just 'TransformCanvas'. Then the namespace implies what it is
  LargeSquare64x64Transformer::TransformCanvas
  RotateTransformer::TransformCanvas
...

hzeller:
In general, looks good to merge; didn't get to it yet (always try to do that on my morning or evening commute :) ). I have one suggestion regarding the inner classes. If you are bored, you can change that, if not, don't worry, I can change that later.

Saij:
No hurry.
Will change that tomorrow.
Also have to implement the transformers into python (and port the samples led-image-viewer and the text sample)

Saij:
OK updated the internal Canvas classes like you suggested.

hzeller:
Alright, merged in two combined commits - one for the transformers, one for the Python changes. Can you verify that things are properly in place as they should be ?

Saij:
Looks good. Thanks for the multiple commits.
Will close this Pull Request now ^^

Issue #52, opened 2015-07-27
Building Python Library


ThomTrab:
Hello,
Thank you very much for all your work, I can not do it but I'm happy to use the python library. But I'm not able any more to build the python library based on the lastest version of the repo on a Raspberry Pi.
Here under the error message:
```
make -C python build
make[1]: Entering directory '/home/pi/tweety_pi/matrix/python'
cython --cplus -o rgbmatrix/core.cpp rgbmatrix/core.pyx

Error compiling Cython file:
------------------------------------------------------------
...
        uint8_t g
        uint8_t b

    cdef cppclass Font:
        Font() except +
        bool LoadFont(const char*)
                               ^
------------------------------------------------------------

rgbmatrix/cppinc.pxd:52:32: Expected ')', found '*'

Error compiling Cython file:
------------------------------------------------------------
...
cimport cppinc

cdef class Canvas:
    cdef cppinc.Canvas *__getCanvas(self) except +
        ^
------------------------------------------------------------

rgbmatrix/core.pxd:4:9: 'Canvas' is not a type identifier

Error compiling Cython file:
------------------------------------------------------------
...

cdef class Canvas:
    cdef cppinc.Canvas *__getCanvas(self) except +

cdef class FrameCanvas(Canvas):
    cdef cppinc.FrameCanvas *__canvas
        ^
------------------------------------------------------------

rgbmatrix/core.pxd:7:9: 'FrameCanvas' is not a type identifier

Error compiling Cython file:
------------------------------------------------------------
...

cdef class FrameCanvas(Canvas):
    cdef cppinc.FrameCanvas *__canvas

cdef class RGBMatrix(Canvas):
    cdef cppinc.RGBMatrix *__matrix
        ^
------------------------------------------------------------

rgbmatrix/core.pxd:10:9: 'RGBMatrix' is not a type identifier

Error compiling Cython file:
------------------------------------------------------------
...
cdef class FrameCanvas(Canvas):
    cdef cppinc.FrameCanvas *__canvas

cdef class RGBMatrix(Canvas):
    cdef cppinc.RGBMatrix *__matrix
    cdef cppinc.GPIO *__gpio        ^
------------------------------------------------------------

rgbmatrix/core.pxd:11:9: 'GPIO' is not a type identifier

Makefile:43: recipe for target 'rgbmatrix/core.cpp' failed
make[1]: *** [rgbmatrix/core.cpp] Error 1
[...]

PR #53, opened 2015-07-27
o Moved generating code from cython to seperate command

Saij:
Discussion in https://github.com/hzeller/rpi-rgb-led-matrix/issues/52


hzeller:
The added dependency compared to the previous version was the problem. I've now solved it slightly differently in 1627ee7a142fd0f6ed3d6370548e1a5e9cf0633f

PR #76, opened 2015-10-24
fix intriguing problem for utilisation with nodejs

zeitungen:
In nodejs utilisation, into the `IsRaspberryPi2` function, the `sscanf` fonction does not set correctly the `mem_size` variable if it does not iniatialize to 0. So the `IsRaspberryPi2` function return `false` and ledmatrix doest not update.

### DIFF
--- a/lib/gpio.cc
+++ b/lib/gpio.cc
@@ -117,7 +117,7 @@ static bool IsRaspberryPi2() {
   buffer[r >= 0 ? r : 0] = '\0';
   close(fd);
   const char *mem_size_key;
-  off_t mem_size;
+  off_t mem_size = 0;
   if ((mem_size_key = strstr(buffer, "mem_size=")) != NULL
       && (sscanf(mem_size_key + strlen("mem_size="), "%lx", &mem_size) == 1)
       && (mem_size == 0x3F000000)) {


hzeller:
Two questions: what do you mean with nodejs use, do you link it to nodejs to create graphics with JavaScript ? (that would be cool).

Second: do you see why this problem you describe happens and what you observed ?

The result of sscanf() is explicitly checked to return the value 1, so it says that it has read the value and filled it into the variable, which initializes it. So I don't see where there could be a problem unless sscanf() disagrees about the byte-width of off_t being long.

zeitungen:
**For the first question**: I want to manage ledmatrix display with javascript function. Here an example what we could do:
```javascript
var LedMatrix = require("node-rpi-rgb-led-matrix").LedMatrix;
var ledmatrix = new LedMatrix(16, 1, 1);
ledmatrix.fill(255, 0, 0);
ledmatrix.setPixel(0, 0, 0, 0, 255);
ledmatrix.clear();
```
and let magic happens ! ;)

I have a start of this [here](https://github.com/zeitungen/node-rpi-rgb-led-matrix) (with an old ledmatrix lib and ugly way to get the ledmatrix lib (not using git submodules)...)

**For the second question**: I don't know why this problem appears but I can describe what I have observed.
If we add some printf to see what happens:
```c
static bool IsRaspberryPi2() {
  // TODO: there must be a better, more robust way. Can we ask the processor ?
  char buffer[2048];
  const int fd = open("/proc/cmdline", O_RDONLY);
  ssize_t r = read(fd, buffer, sizeof(buffer) - 1); // returns all in one read.
  buffer[r >= 0 ? r : 0] = '\0';
  close(fd);
  const char *mem_size_key;
  off_t mem_size;
  // if ((mem_size_key = strstr(buffer, "mem_size=")) != NULL
  //     && (sscanf(mem_size_key + strlen("mem_size="), "%lx", &mem_size) == 1)
  //     && (mem_size == 0x3F000000)) {
  //   return true;
  // }
  if((mem_size_key = strstr(buffer, "mem_size")) != NULL) {
     int res = sscanf(mem_size_key + strlen("mem_size="), "%lx", &mem_size);
     printf("res=%d mem_size=%llu (mem_size == 0x3F000000)=%d\n",
                res, mem_size, (mem_size == 0x3F000000));
     return (res == 1 && mem_size == 0x3F000000);
  }
  return false;
}
```
Here the javascript code for the test:
```javascript
var LedMatrix = require("node-rpi-rgb-led-matrix").LedMatrix;
var ledmatrix;
for(var i=0; i<5; i++) {
   ledmatrix = new LedMatrix(16, 1, 1);
   console.log();
}
```
In my C++ binding code when `new LedMatrix` is called, I do:
```C
assert(io.Init());
matrix = new RGBMatrix(&io, rows, chained_displays, parallel_displays);	
matrix->set_luminance_correct(true);
```

Here the console result:
```
res=1 mem_size=8571359632595877888 (mem_size == 0x3F000000)=0
res=1 mem_size=5351931904 (mem_size == 0x3F000000)=0
res=1 mem_size=9149251004145860608 (mem_size == 0x3F000000)=0

res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=8563499034990346240 (mem_size == 0x3F000000)=0
res=1 mem_size=9149251004145860608 (mem_size == 0x3F000000)=0

res=1 mem_size=9149251158764683264 (mem_size == 0x3F000000)=0
res=1 mem_size=8563499034990346240 (mem_size == 0x3F000000)=0
res=1 mem_size=140270935704862720 (mem_size == 0x3F000000)=0

res=1 mem_size=4834427507400769536 (mem_size == 0x3F000000)=0
res=1 mem_size=8563499034990346240 (mem_size == 0x3F000000)=0
res=1 mem_size=140270935704862720 (mem_size == 0x3F000000)=0

res=1 mem_size=4834427507400769536 (mem_size == 0x3F000000)=0
res=1 mem_size=8563499034990346240 (mem_size == 0x3F000000)=0
res=1 mem_size=140270935704862720 (mem_size == 0x3F000000)=0
```
It works one time (I can play with the ledmatrix), in the second call. But it looks like it works only in the second call.

And when I do `off_t mem_size = 0;` instead of `off_t mem_size;`, I got this:
```
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1

res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1

res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1

res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1

res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
res=1 mem_size=1056964608 (mem_size == 0x3F000000)=1
```

Very interessing, no ?

I hope this is fully understandable ;)

hzeller:
Ah, ok. So it looks like that on your platform, off_t is actually a 64 bit type, while I have only seen 32 bit types so far on Raspberry Pi (on my Pi2, sizeof(off_t) == 4). The sscanf() pattern %lx would only read 32 ... so this would explain why the remaining bits are garbage.

If you print the values you get, e.g. 4834427507400769536 in hex, you see that te remaining bits are actually sound, while the top bits contain garbage 43175634_3f000000

My assumption is, that the build-environment in which you link this library actually sets off_t to be 64 bits, so it fails in this context.

Anyway, the right solution is to fix the type to a known size, so that sscanf() works predictably; I have submitted change 333397e81565d3ef25709da42646e8d676f6d8a4 which should fix this. Can you try ?

The NodeJS port is awesome! I am very excited to see this happen. Let me know when it is ready so that I can add a reference in the README.md.

zeitungen:
Your new modification works !!
Thanks !

I keep you inform when I have a nodejs ledmatrix module ready to work.

zeitungen:
I'm sorry, I've been so exiting and I go too fast when I test.
With `uint64_t` it doesn't work. But with `uint32_t` it's fine.

hzeller:
Even with %llx in the sscanf() option ? That would mean that the implementation sscanf() is actually broken.

zeitungen:
It's an error from me again ...
Because I use old led matrix library code in my nodejs experimentation, I manually merged the modification. And I was very inattentive ...
I have forgotten the `%llx`. With it, it works fine.
Thansk again.

hzeller:
Since we only need 32 bit anyway, I've changed that in commit 
https://github.com/hzeller/rpi-rgb-led-matrix/commit/9fda355ee847fe963d2a7770c9a6458b8c657b20

Please try again.

hzeller:
Ah, looks like I was too fast with the change. Anyway, doesn't matter :)

Probably time for you to switch to the submodule ...

hzeller:
(I have undone the last commit as it wasn't necessary)

zeitungen:
here the nodejs module available on npm: https://www.npmjs.com/package/node-rpi-rgb-led-matrix
here the github repo: https://github.com/zeitungen/node-rpi-rgb-led-matrix

I'm not the only started to bind your library to nodejs:
* https://github.com/Rob-ot/node-rpi-rgb-led-matrix
* https://www.npmjs.com/package/rpi-rgb-led-matrix

But I don't try this binding.

Have fun with this all nodejs stuff ;)

PR #88, opened 2015-11-21
Added HalfTransformer

lanrat:
`HalfTransformer` wraps the canvas exposing only the top half to work with my 16x64 LED display.

Using this API with my display required me to specify `chain=2` and `rows=32` in order for all pixels to be addressable. This however leaves the bottom-half of the canvas addressing non-existent pixels. Using this transformer fixes that.

### DIFF
--- a/include/transformer.h
+++ b/include/transformer.h
@@ -82,6 +82,19 @@ private:
   TransformCanvas *const canvas_;
 };
 
+class HalfTransformer : public CanvasTransformer {
+public:
+  HalfTransformer();
+  virtual ~HalfTransformer();
+
+  virtual Canvas *Transform(Canvas *output);
+
+private:
+  class TransformCanvas;
+
+  TransformCanvas *const canvas_;
+};
+
 } // namespace rgb_matrix
 
 #endif // RPI_TRANSFORMER_H
--- a/lib/transformer.cc
+++ b/lib/transformer.cc
@@ -224,4 +224,68 @@ Canvas *LargeSquare64x64Transformer::Transform(Canvas *output) {
   return canvas_;
 }
 
+/***********************************/
+/* Half Transformer Canvas */
+/***********************************/
+class HalfTransformer::TransformCanvas : public Canvas {
+public:
+  TransformCanvas() : delegatee_(NULL) {}
+
+  void SetDelegatee(Canvas* delegatee);
+
+  virtual void Clear();
+  virtual void Fill(uint8_t red, uint8_t green, uint8_t blue);
+  virtual int width() const;
+  virtual int height() const;
+  virtual void SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue);
+
+private:
+  Canvas *delegatee_;
+};
+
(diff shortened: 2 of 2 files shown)


hzeller:
Can you add a little bit of description comment for the class ? It might not be immediately obvious what the class does. Essentially you support single-line multiplexing instead of common double-line multiplexing.

lib/transformer.cc
```diff
@@ -224,4 +224,68 @@ Canvas *LargeSquare64x64Transformer::Transform(Canvas *output) {
...
+int HalfTransformer::TransformCanvas::width() const {
+  return delegatee_->width();
+}
+
+int HalfTransformer::TransformCanvas::height() const {
+  return delegatee_->height() / 2;
+}
+
+void HalfTransformer::TransformCanvas::SetPixel(int x, int y, uint8_t red, uint8_t green, uint8_t blue) {
+  delegatee_->SetPixel(x, y, red, green, blue);
```
hzeller: mmh, this looks like it will only work with one chain.
If you have one chain (e.g. your 16 pixel high display), then this works.

If you have three of these with `-P3`, then you actually need to do it a little more complicated by mapping 
  * 0..15 -> 0..15  (what you are doing now)
  * 16..31 -> 32..47 (the second panel, which 'thinks' it is from 32..64, but is actually just starting at line 16..)
  * 32..47 -> 48..63 (dito, same sentiment).

include/transformer.h
```diff
@@ -82,6 +82,19 @@ class LargeSquare64x64Transformer : public CanvasTransformer {
   TransformCanvas *const canvas_;
 };
 
+class HalfTransformer : public CanvasTransformer {
+public:
+  HalfTransformer();
```
hzeller: This now only works for 32->16 pixel mapping (see below my SetPixel() comment).

Since the transformer per-se does not know if it is called in a 32 or 16 context (consider a 8 pixel high display, that is behaving like 1/2 a 16 display), this should probably have a constructor parameter.

HalfTransformer::HalfTransformer(int half_height) ?

Then, when you get a delegatee, you can assert(canvas->height() % (2*half_height) == 0);

lanrat:
I can add these checks and documentation if need be, but I'd like to first try to understand the underlying issue I'm having with my matrix.

My matrix is 16x64. When I use this API with `rows=16` and `chain=1` the API claims to have a 32x16  canvas and outputs this when printing "1234567890":
r16c1

When setting `rows=16` and `chain=2` I get a 64x16 canvas, which is correct, but the lower half does not light up.
r16c2


When setting `rows=32` and `chain=1` I get a 32x32 canvas and this:
r32c1

And finally when setting `rows=32` and `chain=2` I get a 64x32 canvas, which is twice as large as my display but I am able to control every LED and then some.
r32c2

I'm guessing that there is some other underlying issue here that my transformer only treats the symptom of, but I don't know enough to be sure.

Are there any other possible settings or options I should try adjusting to see if this can work without the transformer?

lanrat:
Closing this pull request in favor of a better solution in issue #96.

PR #98, opened 2015-12-02
Added support for background colors on DrawText and DrawGlyph

lanrat:
This is particularly useful when updating text over itself, such as a clock.


hzeller:
Thanks! Merged.

PR #113, opened 2016-01-22
ENH: -f flag to flip botton and up in large 64x64 display

yarikoptic:
I have 4 32x32 panels but no long DB9 cables so while arranging them in a square display I couldn't do that long junction as in https://github.com/hzeller/rpi-rgb-led-matrix/raw/master/img/chained-64x64.jpg -- I had to position "right" two pannels to the left which is effectively flipping top and bottom rows from as "-L" arrangement does.  So I have added additional FlipTopBottomTransformer and -f cmdline option to make use of it.

### DIFF
--- a/demo-main.cc
+++ b/demo-main.cc
@@ -1016,6 +1016,7 @@ static int usage(const char *progname) {
           "Default: 1\n"
           "\t-c <chained>  : Daisy-chained boards. Default: 1.\n"
           "\t-L            : 'Large' display, composed out of 4 times 32x32\n"
+          "\t-f            : Flip bottom and up in large 64x64 display\n"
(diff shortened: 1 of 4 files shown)


PacoPeper:
Hello, I have 4 matrix rgb led 32x32 serial connected with Adafruit Hat and need to know how to access each independently via python, to place a different in each text.
Thank you.

aycgit:
@PacoPeper 1: dont ask such questions here, 2: you would be better off putting them all together and displaying diffrent texts at diffrent places.

hzeller:
Not needed anymore these days as there are now `--led-pixel-mapper`; this should be doable with the `Mirror:V` mapper.

PR #122, opened 2016-03-13
Feature/scrambled display

avilleret:
this add a transformer for some scrambled display I found in China
for example, to test one panel : 

```
led-matrix -S -r 8 -c 2
```

each panel contains 2 chains interleaved (thus the name scrambled)
examples should work


hzeller:
Thanks! A few comments before we can proceed

 * Can you base your change on a latest version of the library ? Looks like it is based on some older
version, so the merge does not make much sense (e.g. the if (rows != 8 && rows != 16 && rows != 32) in demo-main.cc already has been there for a while).
 * The pixel-mapping program looks like some test program while you were working on the project, but it should not necessarily be part of this pull request.
 * Can you find a more descriptive name for the Scrambled32x16Transformer ? Scrambled does not really indicate what it does, Striped4Multiplexing32x16Transformer or something is probably more appropriate.
 * What happens if you have multiple displays added in parallel ? Your current set-up assume exactly one chain (and so you set the height to a fixed 16). But if there are multiple in parallel, the height can be multiple of 16. We should make the general case work of multiple of these displays connected in parallel.

avilleret:
thanks for reviewing, I merged your master and fix other points too, feel free to merge or ask me some more changes

hzeller:
(sorry for the delay, I am on vacation right now, will come back to this review as soon as I am back)

adamwolf:
I had to support one of these displays recently.  I merged master into this work, fixed up the one line that needed attention, and now my 16x96 display works with

`sudo ./demo -S --led-chain=6 --led-rows=8 -D7 --led-no-hardware-pulse --led-gpio-mapping=adafruit-hat`

I haven't tackled any of the displays in parallel work from above.  If possible, I'd like to extend this work to get it accepted.  Please let me know if this is feasible.

You can see my update here: https://github.com/adamwolf/rpi-rgb-led-matrix/tree/stripedmultiplexing

hzeller:
There is now `--led-multiplexing` option that allows to choose between different multiplexing variants seen 'in the field'. Can you try if your display now works with the code from head ?

adamwolf:
The installation is finished, but if I need one again or have to do repairs
or modifications I can test then.

Thanks!

PR #139, opened 2016-05-14
Automatic cross-compilation for arm-linux-gnueabihf

encounter:
A couple of changes I've been using in my project to cross-compile the library instead of building it on the device itself.

If the automatic detection isn't desired, I can resubmit with only the first commit so that the library can be used in the context of cross-compilation without any modifications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,10 @@
+TARGET_SYS?=arm-linux-gnueabihf
+SYS:=$(shell $(CXX) -dumpmachine)
+ifneq ($(SYS),$(TARGET_SYS))
+	CXX:=$(TARGET_SYS)-c++
+	AR:=$(TARGET_SYS)-ar
+endif
+
 CXXFLAGS=-Wall -O3 -g
 OBJECTS=demo-main.o minimal-example.o text-example.o led-image-viewer.o
 BINARIES=led-matrix minimal-example text-example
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,3 +1,10 @@
+TARGET_SYS?=arm-linux-gnueabihf
```
hzeller: So I would leave the default target probably an empty string (with a comment what to set it to for cross-compilation (or the environment variable)). Then below making sure that the if-statement below is not executed when $(TARGET_SYS) is an empty string.

Makefile
```diff
@@ -1,3 +1,10 @@
...
+SYS:=$(shell $(CXX) -dumpmachine)
+ifneq ($(SYS),$(TARGET_SYS))
+	CXX:=$(TARGET_SYS)-c++
```
hzeller: Wouldn't this be working better with `$(TARGET_SYS)-g++` ? On my debian installation, I find a `arm-linux-gnueabihf-g++`, but not `arm-linux-gnueabihf-c++` for instance.

hzeller:
Nice! But the full-automatic is typically not what people want: if you just want to do compile tests on your regular machine, it will not work at all unless you have the cross compiler installed.

Maybe just make it half-automatic: If `TARGET_SYS` is not an empty string, then assemble the `CXX:=$(TARGET_SYS)-g++` (and AR), but if it is not set, then just use whatever default compiler is set for CXX.
That way, we don't need the `-dumpmachine` trick either ...

(I quite often actually just compile the library on my regular notebook - I can't send stuff to the matrix, but testing if things compile can be done with the default compiler).

Duality4Y:
oh nice i just edited the makefiles myself to cross compile :)

encounter:
I haven't had the chance to work on this again, so I'm going to close this for now.

Issue #181, opened 2016-07-21
simple zeromq listener


depili:
I have written a very simple zeromq tcp listener for my matrix. The c++ code is in this gist: https://gist.github.com/depili/66cba59a1bbf15d73bc0efb53a7c7401

If you like you can package it with the library, it needs the zmq library and headers. zeromq or 0mq is a pretty nice and simple communication library with bindings for most languages. See http://zguide.zeromq.org/

I use this for generating the bitmaps on a proper pc with ruby and driving the matrix itself with this c++ code on a rpi3.

PR #182, opened 2016-07-22
Zmq server

depili:
This pull request adds a very simple zeromq server to the library examples. This is a refinement of the code presented in issue #181

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 CXXFLAGS=-Wall -O3 -g
-OBJECTS=demo-main.o minimal-example.o text-example.o led-image-viewer.o
+OBJECTS=demo-main.o minimal-example.o text-example.o led-image-viewer.o simple-udp.o zmq.o
 BINARIES=led-matrix minimal-example text-example
 ALL_BINARIES=$(BINARIES) led-image-viewer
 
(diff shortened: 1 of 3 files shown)


hzeller:
mmh, this looks highly specialized to your use-case. Why a complicated zmq socket that requires external dependencies instead of just a simple TCP socket protocol. Your program would be about as short without all the dependency overhead.

depili:
ZMQ deals with breaking up the data into multiple packets if the payload is too large and has bindings for just about every language out there so its pretty universal protocol.

depili:
And one of the reasons why I went with zeromq instead of a simpler udp implementation is that I ran into OSX's udp packet size limit when trying to send 32x128 pixels worth of rgb data.

akatrevorjay:
This is great, thank you!

aycgit:
@depili the idea is great but before this can be merged i suggest you atleast add a minimal api for sending text/images

hzeller:
This is just the library to control the underlying display; any server code to connect to it over the network should be done separately anyway (I suggest by using this library here as a sub-module).

This library does intentionally not suggest any particular way how to interact with the display over the network (in particular not with something requiring an external library like zmq), as this would be the realm of application code. So, this pull request can not be accepted here, but I leave it here for people to draw inspirations.

There are already various examples of other server protocols; for instance
  * [FlaschenTaschen](https://github.com/hzeller/flaschen-taschen/tree/master/server#rgb-matrix-panel-display) provides a layer-composite display over the networks using UDP. Also comes with utitilities to send text or images. The VLC player natively supports FlaschenTaschen to show videos.
  * [PixelPusher](https://github.com/hzeller/rpi-matrix-pixelpusher) implements the PixelPusher protocol, common in certain lighting applications. Supported by Processing and LedLabs.
  * Or here is an old gist I made as an example with a simple [UDP server](https://gist.github.com/hzeller/9ece7d7150c6d28ed1825eee85b5d243)  (if at all, something like this might be considered to be included as an example, as it does not need any additional dependencies).

Knifa:
Hope it's OK to add this as a comment (since this is being left open for people to find look) --- if folk are really looking for a ZMQ-based frame server, I've spun up a project for this here. There's a Docker image available so it's ready to go. I found this to be a much easier paradigm to deal with rather than using the bindings.

https://github.com/Knifa/led-matrix-zmq-server

hzeller:
I've mentioned your implementation in the README now @Knifa so that people can find it there.

PR #186, opened 2016-07-30
Add python 3 support

codingjoe:
I made a tine change to the script to allow absolute imports and added some documentation on how to install it on python 3.4.


hzeller:
Thank you! Merged.
Maybe @Saij (original contributor of the Python addition) can also have a quick look (I personally don't know much about Python).

PR #265, opened 2017-01-05
Code cleanup and refactoring of Python samples

Hopiu:
I tried to refactor the python samples to be a bit more consistent. I moved the -t argument from SampleBase to RunText as it only effects this example. I let SampleBase inherit from object instead of ArgumentParser, as I didn't understood why it did so.
And finally I tried to clean up the code to meet some python coding styles.

### DIFF
--- a/python/samples/graphics.py
+++ b/python/samples/graphics.py
@@ -1,14 +1,15 @@
 #!/usr/bin/env python
 from samplebase import SampleBase
 from rgbmatrix import graphics
-import time;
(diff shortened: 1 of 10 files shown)


hzeller:
Thanks, these changes look good to me, but I am not so much a Python person to know about the Python style and stuff.
I am adding @Saij and @onebeartoe to have a look and comment if they like.
If things are fine I am happy to merge.

PR #268, opened 2017-01-09
Python matrix options

danielvy:
Python - Added support for creating RGBMatrix with RGBMatrixOptions

- Added the structs for RGBMatrix::Options and RuntimeOptions with Python wrapper RGBMatrixOptions
- Added CreateMatrixFromOptions()
- Modified the RGBMatrix constructor to accept optional RGBMatrixOptions and call CreateMatrixFromOptions()
- Modified samplebase.py to create RGBMatrix with RGBMatrixOptions. Also added the more significant options as additional command line arguments
- Updated the readme.md


python/rgbmatrix/core.pyx
```diff
@@ -83,18 +83,100 @@ cdef class FrameCanvas(Canvas):
         def __get__(self): return (<cppinc.FrameCanvas*>self.__getCanvas()).pwmbits()
         def __set__(self, pwmBits): (<cppinc.FrameCanvas*>self.__getCanvas()).SetPWMBits(pwmBits)
 
+cdef class RGBMatrixOptions:
+    def __cinit__(self):
```
hzeller: There should be either the constructor or some InitFromFlags() that takes argc and argv and then calls the underlying rgb_matrix::ParseOptionsFromFlags().

Best: If we have global access to the raw argv/argc in the constructor, we should probably do that there, so that the defaults are automatically set from what the user gave on the command line.

This way, options don't have to be manually added to the samplebase.py and things are always consistent with the C++ demos, otherwise this is a maintenance nightmare, because the C++ library changes relatively often and the Python lib only whenever someone knowledgeable in Python contributes a patch.

python/samples/samplebase.py
```diff
@@ -18,6 +18,10 @@ def __init__(self, *args, **kwargs):
         self.parser.add_argument("-p", "--pwmbits", action="store", help="Bits used for PWM. Something between 1..11. Default: 11", default=11, type=int)
         self.parser.add_argument("-l", "--luminance", action="store_true", help="Don't do luminance correction (CIE1931)")
         self.parser.add_argument("-b", "--brightness", action="store", help="Sets brightness level. Default: 100. Range: 1..100", default=100, type=int)
+        self.parser.add_argument("-m", "--hardware_mapping", help="Hardware Mapping: regular, adafruit-hat, adafruit-hat-pwm" , choices=['regular', 'adafruit-hat', 'adafruit-hat-pwm'], type=str)
```
hzeller: When we have the argument parsing in the RGBMatrixOptions class, we don't need it here (also, it is confusing: are all named differently from the C++ options, which will be a bad user experience and will trigger questions on the forum. So they should be named `--led-gpio-mapping`, `--led-pwm-lsb-nanoseconds`, etc. With the `--led-` prefix and dashes instead of underscores; but anyway, that will happen automatically if we use the standard flags parsing code).

python/rgbmatrix/core.pyx
danielvy: The problem is how to split the args between Python and the C++ lib. Because sample subclasses can (and do) add additional command line args. 

Theoretically, it is possible to tell Python's arg parser to split the args into the "known" args (those added to Python's arg parser) and additional ones that were not defined (which will be passed to C++). I am not sure how well it works. Some code needs to handle the "-h/--help" switch to print the command line. I need to see if this is possible to split those between Python's parser and C++ and it will be inconsistent.

I really think this is a very minor issue since most users will only use the demo briefly before writing their own code that uses the classes directly. 

And in any case, the situation in this pull request is not worse than before, so I don't think this should hold the pull requests.

I will look at how/if this can be done later.

python/rgbmatrix/core.pyx
```diff
@@ -83,18 +83,100 @@ cdef class FrameCanvas(Canvas):
...
+
 
 cdef class RGBMatrix(Canvas):
-    def __cinit__(self, int rows, int chains = 1, int parallel = 1):
-        # TODO(Saij): this should initialize an RGBMatrix::Options and
-        # RuntimeOptions, then call CreateMatrixFromOptions() instead of the
-        # cppinc.RGBMatrix() constructor directly. No __gpio needed anymore.
-        # The options allow to set more things, so they should probably be
-        # available as named parameters in Python ?
+    def __cinit__(self, int rows = 0, int chains = 0, int parallel = 0,
```
danielvy: Yes, it will use the Options defaults: 32,1, 1

python/rgbmatrix/core.pyx
```diff
@@ -83,18 +83,100 @@ cdef class FrameCanvas(Canvas):
...
+            if rows > 0:
+                options.rows = rows
+            if chains > 0:
+                options.chain_length = chains
+            if parallel > 0:
+                options.parallel = parallel
+        # If RGBMatrixOptions provided with also any of rows/chains/parallel
+        # use either RGBMatrixOptions OR rows/chains/parallel
+        elif options != None and (rows > 0 or chains > 0 or parallel > 0):
+            raise Exception("Providing RGBMatrixOptions with rows/chains/parallel parameters not supported.")
```
danielvy: Theoretically, I disagree. Passing duplicate parameters with unclear precedence is not a good API form, imho.

But no problem, I will change this as requested.

python/rgbmatrix/core.pyx
```diff
@@ -83,18 +83,100 @@ cdef class FrameCanvas(Canvas):
...
+
+
+        # For backward compatibility?
         self.__gpio = new cppinc.GPIO()
```
danielvy: I left it for backward compatibility, it case someone uses this to read the GPIO value.

python/samples/samplebase.py
danielvy: I used the name as they appear in the RGBMatrix:Options header. No problem to change. But this will also change the already existing argument names.

python/rgbmatrix/core.pyx
hzeller: So if it is hard to call the C++ parsing, maybe we then should do it 'the Python way'. So I don't know what that would be, but I could imagine, that there could be an Init() function for the RGBMatrixOptions that gets a self.parser or something ?
How is it usually done in Python if submodules need to deal with their own options ?

That then can also cover the `--help/-h` case automatically.

Then the implementation should be in RGBMatrixOptions, so it is in one place (and easy to add if there are new options). That way, we don't have them in samplebase.py which is probably not what people would like to use.

python/samples/samplebase.py
hzeller: (as suggested in my comment above, we should move these options probably into the implementation of RGBMatrixOptions, then this struct deals with the properties and its own flags, so everything is self-contained. Then some call to options.InitFromFlags(self.parser) or something.)

python/rgbmatrix/core.pyx
hzeller: mmh, but doesnt the double-underscore mean that this is really private and nobody should have accessed it ? I'd just remove it (also double-initialization could or could not create trouble with the hardware).

python/rgbmatrix/core.pyx
danielvy: The Python way is what the code does now. Adding the arguments definitions to the arg parser and calling the parse function. Sub modules can be given the opportunity to add their own arguments before the actual parsing and then have access to the args instance to retrieve their values.

it's probably not hard to call the C++ parsing (haven't tried). It's that there are few entities that need to deal with this args, set their switch names and display their help.

I don't think RGBMatrixOptions should know or care about args parsing at all, it's a **library class** and args parsing the **application** logic (and it won't solve the original issue anyway). 

I will try to play with it and see if this can be solved between the SampleBase class and the C++ parsing code and that they can play together nicely. But as I said, it's not a regression from the original code and can be treated separately.

python/rgbmatrix/core.pyx
danielvy: Will remove it then.

python/rgbmatrix/cppinc.pxd
```diff
@@ -40,6 +40,37 @@ cdef extern from "led-matrix.h" namespace "rgb_matrix":
...
+      int daemon
+      int drop_privileges
+
+
+    RGBMatrix *CreateMatrixFromOptions(Options &options, RuntimeOptions runtime_options)
+
+
+
+cdef extern from "led-matrix.h" namespace "rgb_matrix::RGBMatrix":
+    cdef struct Options:
```
danielvy: Re ordering is not a problem. Adding or removing options to the Options struct in C++ requires adding/removing them to the Python definition and recompiling the Python library.

python/rgbmatrix/core.pyx
hzeller: I do think it would be very neat if the RGBMatrixOptions have its own argument parsing for things it is interested in. These are very specific as `--led-rows` or `--led-pwm-lsb-nanoseconds`. Of course, it only works well if Python has a standard support to do these kind of things; it sounds like with the argument parsing thing build into Python it is possible to delegate part of it to other subsystems (but then again, I am not a Python expert. In C, it is a bit awkard, but it commonly done this way for complicated subsystems that pick out arguments).

So the **library** should provide and deal with command line options it is responsible for. That way, the application also does not have to deal with new options that are added over time. The flag namespace starting with `--led-` makes it not clash.

The **application** should deal with options that are relevant for its application doing, such as `--imagepath`.

If the application is interested in providing all the options by themself and not call the RGBMatrixOption matrix parsing code, they can do that as well. But from a using-an-API point of view it is nicer if there is a default way to do this, without having to add the same options every single time.

Look at the current example: they are all deriving from samplebase, mostly because they all have to deal with parsing the options. If they could just call `options.InitFromFlags(self.parser)`, this would be much easier to deal with for someone who wants to quickly see results.

python/samples/samplebase.py
hzeller: Changing the existing long options to the new names (`--parallel` -> `--led-parallel` or `---gpio_slowdown` to `--led-slowdown-gpio`) is no problem for the sake of consistency.

hzeller:
Looks pretty good already, now providing the matrix relevant flags to the RGBMatrixOptions. Did you find if it was possible to use the C++ call ? As I said, I am as well happy if we can pass in an ArgumentParser to some RGBMatrixOptions or whatever standard way there is to have subsystems deal with command line options.

hzeller:
Any news here ? If you want me to work on the flags, then let me know and I can merge it now and add the flag-parsing later the way I envision it.

danielvy:
Haven't looked yet into the issue of sharing the args between the Python samples and C++ library, sorry.

I updated all the argument names to be consistent with the C++ demo.

I think you can merge this now. The argument names are consistent and it allows using all available options from Python, which is an improvement over what we had before without any regressions or breaking backward compatibility.

Hopiu:
@hzeller Any chance to get this merged soon?

hzeller:
Yes, I have been a bit busy, but I try to merge this soon.

hzeller:
Alright, this is merged now into mainline.
Thanks @danielvy !

Barnabe22:
Hello. I'am a beginner and I want to run the python  codes of this project. I have a 32x16 RGBmatrix LED panel. But I don't have the HAT for the matrix and i've done the standard wiring by connecting my rapsberry pi 3 pins to the HUB 75 pins. All is right until here. But the problem is when I run for example:
sudo python runtext.py
Nothing appears on the rgbmatrix LED panel. I don't know why. Will someone clarify something to me please ?

hzeller:
Please don't hijack unrelated issues, just open a new issue with your question.

Before the Python, did you try run the demo program in `examples-api-use/` @Barnabe22 ?

Barnabe22:
Sorry

PR #288, opened 2017-02-18
Added brightness property to FrameCanvas in python.

partofthething:
This is required during double-buffering or else you get crazy flicker
between the two different brightnesses.


hzeller:
Thanks, merged!

PR #381, opened 2017-08-20
added minimal c# wrapper dll and c# example project

digital-synapse:
not sure if you were looking for additional language bindings but here is the start to the C api wrapper with enough support to run the example project in c# on mono


hzeller:
Cool thanks!

Before we can do the merging, can you try to minimize that ?
  * the gitignore blew up significantly in size, but I suspect there are only really a handful new file types if at all that need to be added.
  * Please don't use visual studio and its project file (it is not open source, so it would force people to use a proprietary tool which they might not have access to), but instead write a Makefile that does the same thing. It should entirely be possible to built directly on the Pi with a Makefile. Also, you'll see that simplifies everything significantly.

digital-synapse:
sure i can add a makefile and reduce the gitignore. not a problem, but are you sure you want to exclude the project file? VS is by far the most common IDE for c#

hzeller:
Yes, Makefile is enough. You reduce a complexity to a few lines that is easy to understand and read.

If someone then wants to use the resulting library in Visual Studio they are still free to integrate it in their build system.

hzeller:
Also, the assembly info files don't really look like they are necessary ?

c#/RGBLedMatrix.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace rpi_rgb_led_matrix_sharp
+{
+    public class RGBLedMatrixSharp
+    {
+        [DllImport("librgbmatrix.so")]
+        public static extern IntPtr led_matrix_create(int rows, int chained, int parallel);
```
hzeller: This IntPtr looks like it is not typed, so it is easy to accidentally use wrong.
Is there a way that you can create dummy types so that the pointers that represent a Matrix and a Canvas are different ? I don't know who that is done is CS, so I let it up to you to choose the best way to do that.

c#/RGBLedMatrix.cs
```diff
@@ -0,0 +1,36 @@
...
+        public static extern void led_canvas_set_pixel(IntPtr canvas, int x, int y, byte r, byte g, byte b);
+
+        [DllImport("librgbmatrix.so")]
+        public static extern void led_canvas_clear(IntPtr canvas);
+
+        [DllImport("librgbmatrix.so")]
+        public static extern void led_canvas_fill(IntPtr canvas, byte r, byte g, byte b);
+
+        [DllImport("librgbmatrix.so")]
+        public static extern IntPtr led_matrix_create_offscreen_canvas(IntPtr matrix);
```
hzeller: Aren't we missing led_matrix_get_canvas() ?

digital-synapse:
good to go.
added makefile, removed all the visual studio files and cleaned up the gitignore

digital-synapse:
sorry for all the commits - was trying to clear out the few remaining unneeded files.

hzeller:
Cool, this looks much better. I'll have a closer look tomorrow, its getting night here.

Is there a way to use something like `LD_LIBRARY_PATH` or the equivalent of `-rpath` with csharp programs to help it find the dll ? Right now it looks like everything has to be copied together in a directory for it to work.

Don't worry about all the commits, I'll be squashing them nicely together.

digital-synapse:
As far as i know mono and windows applications always just have their dlls
in the same folder as the executable. However mono does have a concept of a
GAC which may offer somthing like that but i have not used it myself
http://www.mono-project.com/docs/advanced/assemblies-and-the-gac/

digital-synapse:
I looked at adding support for the DrawFont methods but they look quite a bit more involved.

digital-synapse:
Pull request ready for review :)

c#/testapp/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
hzeller: Is this file needed ?

c#/RGBLedMatrix.cs
```diff
@@ -0,0 +1,84 @@
...
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    // TODO: dispose managed state (managed objects).
```
hzeller: So is this something needed and you intend to implement something here, or can this whole if() go ?

c#/Makefile
```diff
@@ -0,0 +1,10 @@
+all:
```
hzeller: This Makefile rule does not make sure that
  (a) the librgbmatrix.so is build. You need to depend on that
  (b) does not do anything depending on the change of the cs files. So you want to depend on each.
  (c) the test program is part of building the library, which is probably what you want to keep independent.

So you probably want to have something like this
```
CSHARP_LIB=RGBLedMatrix.dll
SOURCES=RGBLedCanvas.cs RGBLedMatrix.cs
CSHARP_COMPILER=mcs

RGB_LIBDIR=../lib
RGB_LIBRARY_NAME=rgbmatrix
RGB_LIBRARY=$(RGB_LIBDIR)/lib$(RGB_LIBRARY_NAME).so.1

$(CSHARP_LIB) : $(SOURCES) $(RGB_LIBRARY)
        $(CSHARP_COMPILER) -target:library -out:$@ $(SOURCES)

$(RGB_LIBRARY):
        $(MAKE) -C $(RGB_LIBDIR)
```

for the library, and then an independent target for the test program.

.gitignore
```diff
@@ -1,7 +1,22 @@
...
+*.csproj
+[Dd]ebug/
+[Rr]elease/
+x64/
+x86/
+[Bb]in/
+[Oo]bj/
+[Ll]og/
+.vs/
+AssemblyInfo.cs
```
hzeller: Is this a generated file or why does it show up here ?

.gitignore
```diff
@@ -1,7 +1,22 @@
 *.o
 *.a
 *~
+*.so
 led-image-viewer
 demo
 minimal-example
 text-example
+
+# visual studio
```
hzeller: All these files don't look like they would be generated if the Makefile in this project is being used, so we don't need to mention them here. Probably the only file we need to mention is *.dll

Makefile
```diff
@@ -29,6 +30,9 @@ clean:
 	$(MAKE) -C utils clean
 	$(MAKE) -C $(PYTHON_LIB_DIR) clean
 
+build-csharp:
+	$(MAKE) -C $(CSHARP_LIB_DIR) build
```
hzeller: You don't have a `build` target in your Makefile. So you might either add it there or mention the target `all` that is in the Makefile.

hzeller:
Cool, will have a look later tonight, here is a first round of smallish comments.

Can you add a README.md similar to the one in python/ that describes what needs to be installed (I suspect `mono-mcs`), point out how to run the test program and generalize from there how someone would get started doing their on program (e.g. mention that it is necessary to copy the libraries into the directory of the binary).

To provide more flexibility, it would actually be cool if people could pass something comparable to`struct RGBLedMatrixOptions` as it allows more flexibility than just rows/parallel/chain. But that can as well be a separate pull request.
You then can just `led_matrix_create_from_options()` with an options struct and leave *argc and **argv to NULL.

c#/testapp/App.config
digital-synapse: no it can be removed

c#/RGBLedMatrix.cs
digital-synapse: this was generated the empty if () can be removed

c#/Makefile
digital-synapse: looks good to me. question. why is the library *.so.1 instead of just *.so? I was confused about this

.gitignore
digital-synapse: yes, these folders and files are generated by visual studio. I added them to the gitignore to make it easier for people using VS to contribute. They can be removed from the gitignore if desired.

Makefile
digital-synapse: thanks

digital-synapse:
sure. yeah i think mono-mcs would probably do it for the demo. I'll add a readme with some info.

digital-synapse:
will add the RGBLedMatrixOptions in a seperate pull request.

hzeller:
Thanks!

PR #421, opened 2017-11-03
Added new adapter board for 16x64 LED matrix

clowrey:
These matrix use a non standard pinout - this passive adapter is designed to be very small (low cost)


hzeller:
Nice!

We should probably name this Hub08 adapter or something as it is for the Hub08 pinout. That in fact allows  to re-use more pins and we could do up to 4 panel connectors on a Raspberry Pi with 40 pin GPIO. Do you think you'd like to work on an extended version that allows 4 panels on a single Pi ?

We also should try to get them active with logic-level shifted data lines; the level converters probably save more headaches than the 50 cents they add to the BoM (I am actually considering removing the old passive adapters in my repository to simplify that choice for users :) ).

I've created Issue #531 to track the non-RGB colored panel progress.

clowrey:
Hi! I am pretty busy at this point and ended up switching to using the RGB panels with a passive adapter. 

The ones I bought on AliExpress work fine with passive and let the students easily solder the PCB vs needing surface mount level shifters components.

So I would like to see the passive adapter left in the  repo but obviously with your warnings on their use. 

But if you or someone else sold the adapters premade it obviously makes sense to get the level shifters as the BOM cost is small as you have said.

We have 6x clocks running at a school using 64x128 4mm pitch pixels and your clock demo code ;) looking to get some spare time and integrate the software more with node red to create a message board and weather display system that runs on this hardware but not there yet.

marcmerlin (approved):
I see no harm in merging this even if it's not used by many people.

Please note that there is also this 2 channel passive one:
https://rpi-rgb-led-matrix.discourse.group/t/solved-aliexpress-2-string-passive-adapter-not-working-because-rpi-connector-was-soldered-on-the-wrong-side-of-the-board/892

PR #427, opened 2017-11-10
Library set options

MaitreDede:
Add a way to update (some) matrix options at runtime.


MaitreDede:
Issue https://github.com/hzeller/rpi-rgb-led-matrix/issues/424.
Not yet tested.

MaitreDede:
Brightness tested ok (brightness set to 0 will make it 100%).

hzeller:
So this might be confusing to the user as only few options can even be changed after a matrix has been updated. Even with more documentation this might be confusing.
So I might be more in favor of specific setters for certain fields (such as brightness)

MaitreDede:
Ok, I will do accordingly with latest version.

PR #449, opened 2017-12-04
Imported Configuration File capabilities from adafruit/rpi-fb-matrix

antgiant:
By importing the configuration file capabilities from the adafruit/rpi-fb-matrix repo into the led-image-viewer example, much more complex panel arrangements are supported without requiring any coding. It is also fully backwards compatible.

### DIFF
--- a/utils/Config.cpp
+++ b/utils/Config.cpp
@@ -1,0 +1,135 @@
+// -*- mode: c++; c-basic-offset: 2; indent-tabs-mode: nil; -*-
+// Matrix configuration parsing class implementation.
+// Author: Tony DiCola
+#include <sstream>
+#include <stdexcept>
+#include <vector>
+#include <iostream>
+#include <string>
+
+#include <libconfig.h++>
+
+#include "Config.h"
+
+using namespace std;
+
+// Get value if it exists, otherwise return default.
+static int getWithDefault(const libconfig::Setting& root, const char *key,
+                          int default_value) {
+  return root.exists(key) ? root[key] : default_value;
+}
+
+Config::Config(rgb_matrix::RGBMatrix::Options *options,
+               const string& filename)
+  : _moptions(options),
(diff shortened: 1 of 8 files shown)


marcmerlin:
This looks interesting, but can you explain if this is still needed ? It looks like the U-Mapper will do what this is doing with different config options.

antgiant:
No worries, it has been a while since I thought about this.
Most of the time the existing options can work, and I will admit that the simple U pattern in the example doesn't do this config setup justice. Where this really shines is with more complex wiring set ups. 
Fundamentally, this lets you create a config matrix that is laid out just like your physical screens. Then you can go through and match the order in the config to what you physically wired, including orientation. Want to wire in a spiral or straight across the top row and then a n pattern for the next rows or even just randomly, no problem this config setup can handle them all, while still being reasonably readable.
My original use case was a wiring pattern that did not easily map to one of the existing options. I found this way of doing it much easier than fighting with the options.

marcmerlin:
Ok, that does sound very interesting and needed indeed.
that said, I also just merged this, which I haven't had time to test:
https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/lib#feature-remap-mapper-experimental
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1478

Can you check if this fulfills your need before we add another different way to do things?

antgiant:
It looks to me like what you just merged is just as capable as this. That one works from command line arguments vs this one requiring a config file. Otherwise they look pretty equivalent to me. I find the syntax of that one to be far more confusing but am well aware that is a matter of taste. :-)

marcmerlin:
@antgiant  to be clear, I merged the other one because it would merge as is while yours did not and needed to be rebased. I'm not saying it's better :)
If you have a chance to try it out and let me know if it does the job for you, that would be super appreciated.

Issue #413, opened 2017-10-24
led-image-viewer not working on latest Raspbian Strecht Lite


soyxan:
After installing the required packages:

`sudo apt-get install libgraphicsmagick++-dev libwebp-dev -y
`

If I try to display an animated GIF i get the following runtime error:

`Trouble loading  (Magick: No decode delegate for this image format () reported by magick/constitute.c:1529 (ReadImage))
`

The same GIF file was correctly displayed on Raspbian Jessie Lite.

How can I install the required delegates?

hzeller:
I haven't seen this error message before. But maybe there is some additional image library dependency missing ?

Did you double-check, that the image is readable by everyone ? The image viewer drops permissions to daemon/daemon, so that would need to have read access (you can switch this off with `--led-no-drop-privs` for a quick test if it is permission problems).

soyxan:
The image is readable by everyone and adding your argument to the command-line does not have any effect.
This same setup was working on Raspbian Jessie Lite.
Following the runtime error, I have found this definition of "delegate" in the GraphicsMagick webpage

`Add on third-party libraries and applications which extend the formats supported by GraphicsMagick may be obtained from this directory. Windows users can usually ignore the contents of this directory since all common delegates are included in the Windows source and binaries packages.
`

Even they provide a link to a FTP where you can download those "delegates" but I have no idea how to install them on the Raspberry Pi

[ftp://ftp.graphicsmagick.org/pub/GraphicsMagick/delegates](ftp://ftp.graphicsmagick.org/pub/GraphicsMagick/delegates)

Any idea in how to proceed?

hzeller:
I'd look if there are such delegates available as debian packets.
Also, maybe inform the package owner of that libgraphicsmagick++-dev  debian packet that it does not work as intended.

PR #457, opened 2017-12-14
Extend video-viewer to play videos on loop, all videos passed

t413:
Hey! I also have the issue #413 with the latest raspbian not being able to use libgraphicsmagick– but libavcodec plays gifs quite well. So I extended the video-viewer to be slightly more full-featured.

- video-viewer now includes -r<repeat-secs> to repeat each video for at least x seconds. Great for short GIFs that you'd like to loop a few times.
- video-viewer now loops over all passed video files instead of just the first one.
  - _This may seem like overkill, couldn't one just call video-viewer again in a loop? Yes– however the system loading time to instantiate and initialize the matrix, the avcodec library, etc adds several seconds of blank screen between very short videos. Allowing video-viewer to play each file passed makes it consistent with the image viewer._
- Moved the -R and -L options over to text-example, allowing rotated displays and a U-shaped display just like image and video viewers.
- Added a few more stats to the stderr output in video-viewer with frames decoded, time played, and filename.

All of the changes preserve the exact default behavior of each app. I've tried to keep the same code styling and use consistant option names between apps.

Hope this can be useful to more than just me and my friend (who is already using these additions)!

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -166,7 +166,7 @@ public:
   void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
     const bool is_left_check = (x < panel_cols_/2);
 
-    if((y <= 7) || ((y >= 16) && (y <= 23))){
+    if ((y <= 7) || ((y >= 16) && (y <= 23))){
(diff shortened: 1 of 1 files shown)


himura95:
Thanks you very much !!!

PR #505, opened 2018-02-04
Don't ignore pwm_lsb_nanoseconds when using C API.

rossbacher:
When using the C API `pwm_lsb_nanoseconds` in `RGBLedMatrixOptions` was ignored.


hzeller:
Thanks!

Issue #278, opened 2017-02-07
Multiple pixel lit


agoesssss:
Hi there, is there anyone ever experienced multiple vertical pixel lit when trying to SetPixel single pixel. I'm trying to turn on single pixel but I always got 4 vertical pixel lines. 

Can anyone address this issue? 

Thank you

agoesssss:
![image](https://cloud.githubusercontent.com/assets/25609263/22688681/e9ca92f6-ed5f-11e6-816e-7761607cd777.png)

That is what I got when doing SetPixel(0, 0, 255, 255, 255);

hzeller:
This is interesting, does that happen on any pixel, eg. if you choose one in row 2, does it go from row 2 to 6 ?

This might be a slightly different panel control. CS sounds suspicious (this abbreviation usually means 'Chip Select', used to tell a subsystem that whatever data it gets is supposed for it).
It can be detrimental if that is floating (i.e. not connected). Try to set that to 5V in one experiment and to GND in another (I expect the panel to not show anything in one or the other experiment).

Can you take a good readable picture of the connector on the back of the panel and post here ?

It could also be a different issue in which it does multiplexing a bit differently, which we have to find out.

agoesssss:
I guess that's nothing to do with the chip select that I also suspect earlier, because when I tried to pulse that CS then my system freeze and nothing happened when I set to ground, and then I realized that CS is shorted to the ground. I guess there's issues with multiplexing

these is what I got when doing
matrix_->SetPixel(0, 0, 255, 0, 0);
matrix_->SetPixel(1, 1, 0, 255, 255);
matrix_->SetPixel(2, 2, 0, 0, 255);
![image](https://cloud.githubusercontent.com/assets/25609263/22722093/3e85272a-ede7-11e6-8e71-12246354cca5.png)

matrix_->SetPixel(1, 0, 255, 0, 0);
matrix_->SetPixel(2, 0, 0, 255, 255);
matrix_->SetPixel(3, 0, 0, 0, 255);
![image](https://cloud.githubusercontent.com/assets/25609263/22722128/870da94a-ede7-11e6-97a6-2f470ea6ee79.png)

matrix_->SetPixel(0, 1, 255, 0, 0);
matrix_->SetPixel(0, 2, 0, 255, 255);
matrix_->SetPixel(0, 3, 0, 0, 255);
[...]

PR #509, opened 2018-02-09
Added new RowAddressSetter class and transformer class for 1:4 multiplexing matrix

vapiper:
Added new RowAddressSetter class and transformer class for outdoor matrix described in issue #278


hzeller:
Nice, thanks for your contribution, I'll have a look at it tonight. Looks like you had to figure out two puzzling properties to get your panel working.

Maybe you also would like add a row to the https://github.com/hzeller/rpi-rgb-led-matrix#types-of-displays in the toplevel README to mention the row-address-type for people encountering a similar 1:4 panel ?

And one question for the `ZStripeTransformer`: does it work with multiple chained and parallel displays ? The way the tranformers are set-up currently for the Multiplexing (which I should probably simplify) is that they get the full length display made out of the chain of panels and the height that can depened on the parallel panels. This is why all these transformers have the funny

```
 const int chained_panel  = x / panel_cols_;
 const int parallel_panel = y / panel_rows_;

 const int within_panel_x = x % panel_cols_;
 const int within_panel_y = y % panel_rows_;
```
-thing going on and using that to find which matrix inside the larger panel is used, and then compose it together in the end.

So as I said, I should probably simplify that and pull into the `BasicMultiplexCanvas` so that each of these transformers only has to deal with the individual panel, but for now, you probably have to add the same logic in your transformer.

vapiper:
I have to add in the README information about such matrices, . Please, check my changes. English is not my native language :-).

I made changes in the `ZStripeTransformer` class as you wrote. 
I think it's really an easier way to develop new classes of transformers, because you need to write less code. To understand pixel mapping in this matrix here is a link to the video [X10 pixel mapping](https://youtu.be/r6vtvPwIer0).
I tested the `ZStripeTransformer` with a chain of 2 matrices. It works great. 
img_2377

hzeller:
Thanks!

angelogoncalve:
for the x10 is that?: --led-rows=16 --led-row-addr-type=2 --led-multiplexing=0

and for the q10 is this mode?: --led-rows=16 --led-row-addr-type=2 --led-multiplexing=4

vapiper:
X10 and Q10 is a matrix of the manufacturer Qiangli. X10 and Q10 are the same matrices with different markings. In 2016, the manufacturer called it X10, since 2017  - Q10. 
img_2378
The new classes in this pull request are tested with both matrices. If you have the Q10 or X10 matrix then use the parameters: `--led-rows=16 --led-row-addr-type=2 --led-multiplexing=4`

hzeller:
Note, I have now created a new branch `pixel-mapper`

https://github.com/hzeller/rpi-rgb-led-matrix/tree/pixel-mapper

.. in which writing new MultiplexMappers (the new name) will be simpler. And adding a mapper is only done at one place, the other places in [led-matrix.cc](https://github.com/hzeller/rpi-rgb-led-matrix/blob/pixel-mapper/lib/led-matrix.cc#L188) and [options-initialize.cc](https://github.com/hzeller/rpi-rgb-led-matrix/blob/pixel-mapper/lib/options-initialize.cc#L434) are automatically picking them up; no need to change code in three places.

I've also ported [your ZStripeTransformer](https://github.com/hzeller/rpi-rgb-led-matrix/blob/pixel-mapper/lib/multiplex-mappers.cc#L137). Please check out if it works.

Also let me know if you have improvment suggestions of the new API.

There are still some small things I'd like to do before merging them with master, but this is roughly the new way how to do the same things simpler.

vapiper:
It is a great idea. The code has become easier and clearer at new branch `pixel-mapper`. Create own the `MultiplexMapper` class has become more simply. 
In my opinion, there is confusion in terms in the method `MultiplexMapperBase::MapSinglePanel(int phys_x, int phys_y, int *logic_x, int *logic_y)`, because it must transform from coords of the display (visible) to coords of the matrix (phys).

I have performed the tests at my matrixes. It works fine. Thank you!

At the moment, the API suits me completely. If I will have any new ideas, that I will sure let you know.

hzeller:
Thanks for the feedback; I have now [renamed](https://github.com/hzeller/rpi-rgb-led-matrix/blob/pixel-mapper/lib/multiplex-mappers.cc#L68) the method parameters to `visible_x, visible_y` and `matrix_x, matrix_y`:

```
void MapSinglePanel(int visible_x, int visible_y,
                    int *matrix_x, int *matrix_y) const
```

PR #528, opened 2018-02-22
 Set brightness for all created FrameCanvas instances at runtime

vapiper:
When you use  double or multibuffering, calling the `RGBMatrix::SetBrightness` method changes the brightness of the active frame, other frames are displayed with the previous brightness value.
I made changes in the method `RGBMatrix::SetBrightness`.  Now the brightness is set for all instances of the `FrameCanvas` class. This can be used to automatically adjust the brightness of the matrix at runtime, e.g. from a light sensor.


hzeller:
Thanks, I have mostly a stylistic request before we can merge this.

vapiper:
What did you mean "mostly a stylistic request"?
If I got you right, the method code should be like this.:
```
void RGBMatrix::SetBrightness(uint8_t brightness) {
  for (size_t i = 0; i < created_frames_.size(); ++i) {
    created_frames_[i]->framebuffer()->SetBrightness(brightness);
  }
  params_.brightness = brightness;
}
```

hzeller:
Doh, github does not submit the comment I had pending when I sent the reply. Apparently I had to press another button. Sorry about that.

Yes, that was what I meant...

vapiper:
Ok. I changed the code style.

hzeller:
Thanks, merged!

PR #538, opened 2018-03-02
Add new MultiplexMapper

vapiper:
Add new `MultiplexMapper` based on the modified `ZStripeMultiplexMapper` class. See issue #529


hzeller:
Thanks!

PR #560, opened 2018-03-18
Add demo a demo for reading pixels from stdin

polyfloyd:
This is a simple demo program that reads it's frames from STDIN. This essentially decouples the animations/programs from their output, making development and reuse of animations easier.


hzeller:
Cool! Thanks merged!

PR #581, opened 2018-04-04
New mapper for P10 outdoor panel

aldan4uk:
Added new mapper for P10 outdoor RGB panel based on LEDASIC PR4538DW chip.

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -178,6 +178,80 @@ public:
   }
 };
 
+class P10OutdoorMapper : public MultiplexMapperBase {
+public:
+  P10OutdoorMapper() : MultiplexMapperBase("p10outdoor", 2) {}
+
+  void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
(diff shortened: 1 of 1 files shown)


lib/multiplex-mappers.cc
```diff
@@ -178,6 +178,80 @@ class CoremanMapper : public MultiplexMapperBase {
   }
 };
 
+class P10OutdoorMapper : public MultiplexMapperBase {
+public:
+  P10OutdoorMapper() : MultiplexMapperBase("p10outdoor", 2) {}
```
hzeller: Instead of P10OutdoorMapper (which is essentially all the mappers you find in this file), can you name it according to the characterisitc it does ? Above, there is for instance a mapper that 'snakes' through the pixels, thus its name. Is there anything you could describe what this mapper is doing to give it a nice distinctive name ?

hzeller:
Do you consider fixing the requested changes, or maybe some mapping added in the meantime does what you need ?

marcmerlin:
if this is still desired, please submit an updated patch against master

PR #594, opened 2018-04-19
Add demo method to fill in all pixels in the display

cmagnuson:
I've been testing a HUB12 panel and using this demo method to try and determine/test the matrix transformation. Thought it might be useful to include as part of the standard demo program

### DIFF
--- a/examples-api-use/demo-main.cc
+++ b/examples-api-use/demo-main.cc
@@ -216,6 +216,34 @@ private:
   }
 };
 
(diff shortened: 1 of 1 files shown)


cmagnuson:
Made requested changes

marcmerlin:
apologies that this never got merged. If you're still around, can you resubmit and I'll merge.
this one is way too old and won't merge, so I'll close it to help clean the queue.

PR #657, opened 2018-07-07
Add free-i2c mapping  and rename "U-mapper" to "Snake" with extra lines param

arduino12:
Add "free-i2c" gpio-mapping for using 3 parallel chains among i2c bus.
Add "lines" param to new "Snake" pixel-mapper instead of old "U-mapper",
the default lines value is 2- like the old "U-mapper",
but it can be changed like "Snake:4" for a 4 panels height display..!


marcmerlin:

But generally, yes I support this mapper, it is needed. If you can provide an updated version, I'm in favor of merging it.

marcmerlin:
You can also commit free-i2c separately (if it had been separate I'd have merged it on the spot)

PR #675, opened 2018-08-03
Control Panel

laesskerjonathan:
Web Application to control


davemaster:
Greetings,

Try something like this:

web based rgb led panel screen

PR #770, opened 2019-03-10
fix link to Adafruit section in README.md

Runinho:
The link to the Adafruit section is broken.
In the current state, it links to `#if-you-have-an-adafruit-hat` instead of `#if-you-have-an-adafruit-hat-or-bonnet`.


hzeller:
thanks! The headline changed at some point, but looks like the link was not updated...

PR #777, opened 2019-03-19
Added a link to the Acme Systems adapter

tanzilli:
Acme Systems offers a smaller ready-to-use adapter based on Active-3 board schematic.


marcmerlin:
apologies for dropping this, I just manually merged your change and will drop this one since it conflicts

PR #785, opened 2019-03-26
MirrorPixelMapper with the ability to apply horizontal and/or vertical mirror image

yawor:
Just a simple mapper which can be used to mirror the image on the matrix horizontally and/or vertically. It defaults to a horizontal mirror, but it can be changed with an argument.
Horizontal mirror:
`--led-pixel-mapper=Mirror` or `--led-pixel-mapper=Mirror:h`
Vertical mirror:
`--led-pixel-mapper=Mirror:v`
Both directions (gives the same effect as `Rotate:180`):
`--led-pixel-mapper=Mirror:hv`

### DIFF
--- a/lib/pixel-mapper.cc
+++ b/lib/pixel-mapper.cc
@@ -90,6 +90,53 @@ private:
   int angle_;
 };
 
+class MirrorPixelMapper : public PixelMapper {
+public:
+  MirrorPixelMapper() : direction_(0) {}
+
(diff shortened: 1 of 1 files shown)


hzeller:
Oh, just seeing this. I guess I missed it back then and in the meantime, there is already a mirror option (implemented in 9a8a2c12a278117dfc959a1743e22f200f40502b ).

PR #786, opened 2019-03-26
ji/jihat

jayeye:


lib/Makefile
```diff
@@ -21,10 +21,11 @@ TARGET=librgbmatrix
 #   regular            # Following this project wiring and using these PCBs
 #   adafruit-hat       # If you have a RGB matrix HAT from Adafruit
 #   adafruit-hat-pwm   # If you have an Adafruit HAT with PWM hardware mod.
+#   ji-hat             # https://github.com/jayeye/ji/src/kicad/rpi-rgb-led-matrix-ji
 #   regular-pi1        # If you have an old Pi1 and regular didn't work.
 #   classic            # (deprecated) Classic Pi1/2/. Not used anymore.
 #   classic-pi1        # (deprecated) Classic pinout on Rasperry Pi 1
-HARDWARE_DESC?=regular
+HARDWARE_DESC?=ji-hat
```
hzeller: You probably don't want this default to be changed.

lib/hardware-mapping.c
```diff
@@ -146,6 +146,93 @@ struct HardwareMapping matrix_hardware_mappings[] = {
...
+    .p2_b1         = GPIO_BIT(7),
+    .p2_r2         = GPIO_BIT(8),
+    .p2_g2         = GPIO_BIT(6),
+    .p2_b2         = GPIO_BIT(11),
+  },
+
+  /*
+   * The mapping for JI's hat(s):
+   * https://github.com/jayeye/ji/src/kicad/rpi-rgb-led-matrix-ji
+   * The first batch of the zero adaptor had swapped G1/G2 lines.
```
hzeller: Given that this was possibly just a test-run with only you or a few other having seen the hardware, do you really want to include the buggy version of the adapter in the main rpi-rgb-led-matrix distribution ?

marcmerlin:

PR #851, opened 2019-07-16
Adds support for P10 Coreman Outdoor panels

calcinai:
This patch adds support for outdoor panels like [this one](https://www.aliexpress.com/item/32635788336.html).

The pixel mapping is as follows

Coreman Pixels

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -239,6 +239,23 @@ public:
   }
 };
 
+class P10CoremanMapper : public MultiplexMapperBase {
+public:
+  P10CoremanMapper() : MultiplexMapperBase("P10CoremanMapper", 4) {}
+
+  void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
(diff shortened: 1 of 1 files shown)


calcinai:
So... I've just realised that there's already a Coerman mapper, although it doesn't work for this panel.  Do you have a suggestion for better naming?

angelogoncalve:
In the library produce by Henner Zeller for this type of panel Coreman 16x32 with scan rate 1/4 (outdoor), you can see this type of  pixel mapping is supported:


https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/multiplex-mappers.cc

calcinai:
@angelogoncalve I've tried all of the existing multiplexing modes, and none function correctly.  The one that's closest is actually `Kaler2Scan`.

Can you please give me some more information how I can use an existing mapper with the pixel mapping above?

hzeller:
Merged manually. Sorry for the late response.
I just kept it the name you provided - in the end it is not too critical as people will just go manually through the list until something works :)

PR #858, opened 2019-07-28
fix resize bug.

pokiiio:
Resized images were not used.

### DIFF
--- a/bindings/python/samples/image-viewer.py
+++ b/bindings/python/samples/image-viewer.py
@@ -22,7 +22,7 @@ options.hardware_mapping = 'regular'  # If you have an Adafruit HAT: 'adafruit-h
 matrix = RGBMatrix(options = options)
 
 # Make image fit our screen.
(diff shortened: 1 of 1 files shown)


hzeller:
mmh, what do you try to fix ? The thumbnail() should do exactly what we want, as it keeps the aspect ratio but makes it fit inside the given size.
If you use resize(), you'll force the image to be in the aspect ratio the matrix is, which typically means it will be distorted.

pokiiio:
Oh, you are right.
But if we use a smaller image than matrix, thumbnail() will not expand it. (and I often use the smaller images...) So I changed the code.

Ideally, the code should change resize() and thumbnail() dynamically depending on the image and the matrix sizes. However, in major use case, bigger images will be used.

hzeller:
Yes, for the sample use-case of this script, thumbnail() is better, but people will adapt the code to whatever is needed in their case (like you needed resize()).
But let's keep the default case simple and useful for the common case.

PR #862, opened 2019-07-31
Include more details about the mapping profiles

Eibwen:
TODO still need to add a link to this from the other documentation pages, any thoughts on that?

I was a little confused because I followed a tutorial which used `classic`, and wasn't able to figure that out until looking at the source of the mapping definitions.  Thought having this details explained might be helpful in to others and for debugging

### DIFF
--- a/wiring_detail.md
+++ b/wiring_detail.md
@@ -1,0 +1,83 @@
+# Description
+This library supports a number of GPIO mappings.  The header column of this table can be passed to `--led-gpio-mapping=<name>` to have the library work with that GPIO mappings.
+
+The `pi1` mappings were quite limited on the number of GPIO outputs, so used the pins which hardware I²C needs.  If you use a modern RPi with more GPIO pins that mapping will work, but you cannot use I²C modules along with it.
+
+## HardwareMapping Property Name
+Bold numbers are **physical pins**, everything else is BCM GPIO pins
+|  | regular | adafruit-hat | adafruit-hat-pwm | regular-pi1 | classic | classic-pi1 | **HUB75 physical pin** |
+|---|---|---|---|---|---|---|---|
+| .output_enable | 18 | 4 | 18 | 18 | 27 | 0 \| 2 | **15** |
+| .clock | 17 | 17 | 17 | 17 | 11 | 1 \| 3 | **13** |
+| .strobe | 4 | 21 | 21 | 4 | 4 | 4 | **14** |
+| /* Address lines */ ||||
+| .a | 22 | 22 | 22 | 22 | 7 | 7 | **9** |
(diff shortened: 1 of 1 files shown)


hzeller:
(note, your table does not render properly with the github markdown. Try `grip` to test locally)

But I'd make it simpler to not confuse people too much.
I'd probably give less detail of the mapping here as t might change and it means the same information is in multiple places, but just point out that there are different hardware mappings that the user can use with the flag.

Maybe just a section in `wiring.md` which has a description that there are also different mappings available than the 'regular' one above, and then have a table: Probably a table that describes 'regular', 'adafruit-hat', 'adafruit-hat-pwm' ... etc. is a good idea.

Then, point to the [lib/hardware-mapping.c](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/hardware-mapping.c) file for details if someone is interested or wants to build their own mapping.

If you find an old tutorial that points out the ancient mapping, please tell them to add a huge header that this is outdated and that they should go to https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/wiring.md directly.

Eibwen:
Ah, I would have expected most of the mappings would not change, due to them being based on physical PCBs and all.

Actually in hindsight one goal I was having was debugging the hardware connections I had created back to the software ones.  But then I did end up just using the BCM GPIO numbers.... but ended up writing that all out on paper, and didn't update the table for that.

Would you consider a table of physical pins to physical pins be of similarly not adding much value due to being redundant/confusing?

>Probably a table that describes 'regular', 'adafruit-hat', 'adafruit-hat-pwm' ... etc. is a good idea.

What exactly do you mean by "describes" here?  Similar to my (now fixed) first table?  Or a table form of https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/README.md#changing-parameters-via-command-line-flags with all the possible options included?:

| --led-gpio-mapping |
| --- |
| regular |
| adafruit-hat |
| adafruit-hat-pwm |
| regular-pi1 |
| classic |
| classic-pi1 |

(also where I made my mistake:) I can't recall if I found both that tutorial and your write up, but I believe I went with the ancient tutorial partially because it had the graphic with lines rendered connecting the pins, which at the time I felt was a good double-check for the work I was doing, and didn't check closely enough to realize that it didn't match up with your documentation

Attempted to submit feedback on the classic-rpi1 tutorial, but not certain if it actually worked or not with their interface.

hzeller:
I've now added a table with the GPIO mapping [directly in the wiring.md page](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/wiring.md#alternative-hardware-mappings), but put it in a foldable 'detaill section, so that it is not too prominent. What do you think ?

PR #909, opened 2019-10-16
Adding a still image viewer that works with either a local path or a url.

midorineko:
Image Viewer doesn't set options based on the arguments used in the rest of the application. Things like U-Mapping are impossible to set without manually editing the code and adding an option for U-mapping. 

StillViewer will automatically take a local or web image and display it on a LED Matrix. 
StillViewer will also determine if the image is landscape or portrait and automatically resize and crop it.
It has the latest option flags and supports U-mapping.


midorineko:
It doesn't use  samplebase.py so options have to be manually set.

PR #921, opened 2019-11-03
fixed GpioSlowdown parameter passing

### LINKED ISSUE #906: RGBLedMatrixOptions.GpioSlowdown property has no effect (C# binding)
Setting the `RGBLedMatrixOptions.GpioSlowdown` property has no effect. It always has the default value of 1.

In my opinion, this should never have worked before. While options like rows, cols, chain length, etc are part of `RGBMatrix::Options`, gpio-slowdown is part of `RuntimeOptions` (and is never passed to `led_matrix_create_from_options`).

My current workaround is to hardcode set `DRGB_SLOWDOWN_GPIO` in `lib/MAKEFILE`

daniel-richter:
* _GPIO slowdown_ is part of `RuntimeOptions` (and will never be passed to `led_matrix_create_from_options`)
* pass it as `--led-slowdown-gpio` to `argc`/`argv` manually when invoking `led_matrix_create_from_options`

this commit fixes #906


hzeller:
Thanks!

PR #951, opened 2019-12-28
Swap options parameter for RGBMatrix constructor

DismissedGuy:
This PR makes a small change to the constructor of the `RGBMatrix` class by shifting the kwarg to specify the options to the front.
IMO, this implementation makes more sense since using an `RGBMatrixOptions` object is the main way to specify your layout and other options, and this would allow you to initialize your matrix object like this:
```python
options = RGBMatrixOptions()
options.rows = 64
matrix = RGBMatrix(options)
```
which was previously not possible as doing this would pass your options as the `rows` kwarg and cause a segfault.

### DIFF
--- a/bindings/python/rgbmatrix/core.pyx
+++ b/bindings/python/rgbmatrix/core.pyx
@@ -184,8 +184,8 @@ cdef class RGBMatrixOptions:
         def __set__(self, uint8_t value): self.__runtime_options.drop_privileges = value
 
 cdef class RGBMatrix(Canvas):
-    def __cinit__(self, int rows = 0, int chains = 0, int parallel = 0,
-        RGBMatrixOptions options = None):
+    def __cinit__(self, RGBMatrixOptions options = None,
+                  int rows = 0, int chains = 0, int parallel = 0):
 
         # If RGBMatrixOptions not provided, create defaults and set any optional
         # parameters supplied


hzeller:
mmh, but how would backwards compatibility work ? If someone already uses the old constructor with rows, columns etc. defined, their code will break. This is the reason that the options parameter is at the end to preserve backward compatibility.

Is it possible in Python to have multiple constructors ? One with the row/chain/parallel and one with the options ?

DismissedGuy:
If they specify `rows`, `chains` and `parallel` as kwargs, there shouldn't be a problem. If they don't though, and specify the values as normal positional arguments, you're right, that would break backward compatibility.

It would be possible to have multiple "constructors" by using class methods. That way, you would instantiate the matrix object something like this:
```
options = RGBMatrixOptions()
options.rows = 32

# for row, chain, parallel args
matrix = RGBMatrix(1, 2, 3)  # use your own args

# for options
matrix = RGBMatrix.from_options(options)
```
This would also break backward compatibility though, unless we keep the options param at the back where it used to be to allow usage as currently documented in the README (pass `options` as kwargs to the constructor.)

Another possible solution would be to use [Union types](https://www.python.org/dev/peps/pep-0484/#union-types) and check whether the first parameter is either an int (specify `rows`) or an `RGBMatrixOptions` object, but IMO this solution would be rather ugly. I think it's best to keep the old behaviour but add a classmethod that allows you to pass in the options as a positional argument, rather than a keyword argument.

hzeller:
I think I'd be move in favor or just removing the other parameters and just go with the options.
If we break backwards compatibility, we should use the opportunity to encourage the right way to do things.

DismissedGuy:
Sounds good, I think the most common way to instantiate the object is to use the options anyway.
As part of the change, it would probably be a good idea to explicitly check the argument type to prevent segfaults, since that used to happen when you passed an argument with an invalid type.

If you want to, I could probably make some changes in the next few days, although it might take a bit before I've got time.

hzeller:
Sounds good!

DismissedGuy:
I've made some changes which should theoretically work, however after building from my fork and testing it out, it still seems to be using the old constructor. Not sure why as I've got near-0 experience with python bindings (or lower level languages in general), but I'm guessing it requires modifications to the underlying C++ code as well?

PR #952, opened 2019-12-29
Add panel type binding for Python to support FM6126A chip

mister-scott:
Add panel type binding for Python to support FM6126A chip in response to our need for support on that panel.

See the following commit message for context.
https://github.com/hzeller/rpi-rgb-led-matrix/commit/7160a8ade3dcecabe25ce0de73232b365f7587be

#firstpullrequest
Collaborated on with @greglaw


bindings/python/samples/samplebase.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, *args, **kwargs):
         self.parser.add_argument("--led-pixel-mapper", action="store", help="Apply pixel mappers. e.g \"Rotate:90\"", default="", type=str)
         self.parser.add_argument("--led-row-addr-type", action="store", help="0 = default; 1=AB-addressed panels;2=row direct", default=0, type=int, choices=[0,1,2])
         self.parser.add_argument("--led-multiplexing", action="store", help="Multiplexing type: 0=direct; 1=strip; 2=checker; 3=spiral; 4=ZStripe; 5=ZnMirrorZStripe; 6=coreman; 7=Kaler2Scan; 8=ZStripeUneven (Default: 0)", default=0, type=int)
+        self.parser.add_argument("--led-panel-type", action="store", help="Needed to initialize special panels. Supported: 'FM6126A'", default="FM6126A", type=str)
```
hzeller: I would not set the default to FM6126A. This looks like something you need for your configuration, but for the context of this pull-request, this should be just an empty string.

bindings/python/rgbmatrix/core.cpp
```diff
@@ -1,4 +1,4 @@
-/* Generated by Cython 0.29.6 */
+/* Generated by Cython 0.29.2 */
```
hzeller: If possible, I'd try to keep the same cython version as before (0.29.6), so that the differences are only reflecting the changes due to the new property.

bindings/python/samples/samplebase.py
mister-scott: Corrected per your recommendation. Now default value is "" instead of "FM6126A".

bindings/python/rgbmatrix/core.cpp
mister-scott: Rebuilt using the v0.29.6 of cython.

hzeller:
Thanks! merged.

PR #968, opened 2020-01-19
implemented InversedZStripe

etanremlap:
I'm not the best at coming up with names, so this may not be the best name.

ZStripe looked to be the closest match for me, so I started with that, and added a couple of transformations to get my board working.


hzeller:
Thanks!

Issue #918, opened 2019-10-29
Getting high refresh rates on a 192x128 matrix


carlosalaniz:
I'm driving a 192x128 matrix, however, the RPI can only go up to 60 ~ 100hz when playing an animation.

Chinese controllers can achieve high refresh rates 960hz+, however, these cards do not provide a good interface to be controlled programmatically, I need to be able to control individidual pixels on the matrix, stream video, among other things, that's why I'm looking into this library, anything that you would recommend to accomplish this at high refresh rates?

Thank you!

hzeller:
You need to give a little more detail about your setup.

What are the panels used, how many parallel and chains. How you do the animation? pwm depth?
What are the parameters you give to the library? What is the multiplexing? are you using pwm dither?

Cellgalvano:
@carlosalaniz These chinese controllers use FPGAs to achieve high refresh rates, they are doing the whole refresh in hardware. I do not know if one Pi is fast enough for such a big panel, if it isn't you could look into using a FPGA Hat for the Pi.

marcmerlin:
@carlosalaniz @Cellgalvano   it just so happens that I also setup a 192x128 matrix, and it was ok enough in some setups.
~/rpi-rgb-led-matrix/examples-api-use/demo --led-gpio-mapping=regular --led-rows=64 --led-cols=128 --led-row-addr-type=0 --led-chain=3 --led-show-refresh --led-slowdown-gpio=1 --led-parallel=1 --led-pwm-dither-bits=1 --led-pwm-lsb-nanoseconds=9 --led-pwm-bits=4 -D0
gives 390Hz, but --led-pwm-bits=4 isn't great

~/rpi-rgb-led-matrix/examples-api-use/demo --led-gpio-mapping=regular --led-rows=64 --led-cols=128 --led-row-addr-type=0 --led-chain=1 --led-show-refresh --led-slowdown-gpio=1 --led-parallel=3 --led-pwm-dither-bits=1 --led-pwm-lsb-nanoseconds=9 --led-pwm-bits=7 --led-panel-type=FM6126A -D10
aso gives 350Hz or so but with 7 bits of color (note that I had to patch the library to allow  --led-pwm-lsb-nanoseconds=9 which gives a small boost, not sure if @hzeller  would take a patch to remove the restriction that currently restricts the speed to 50ns).
Note that this 2nd invocation uses the parallel board with 3 outputs.

[...]

PR #971, opened 2020-01-24
Capture performance discussion from issue 918

marcmerlin:
https://github.com/hzeller/rpi-rgb-led-matrix/issues/918

Add performance tips and reasonable limits to expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -685,31 +685,29 @@ so fast to a string of panels before you get flickering due to too low a refresh
 rate (less than 80-100Hz), or before you refresh the panel lines too fast and they
 appear too dim because each line is not displayed long enough before it is turned off.
 
-AB panels also known as outdoor panels, help with performance (you can drive them a bit
-faster and they'll still look bright enough), but as you drive them faster, you'll
-eventually run into the same limits.  
-Ultimately a Raspberri Pi3 is fast enough to push pixels as quickly as panels can accept them
-(actually, not quite an rPi4 goes a bit faster even with the required --led-slowdown-gpio=2).  
-Using an FPGA solution could work around some slight inefficiencies but ultimately run into 
-similar limits.  
-A general rule of thumb is that running 128x128 on a single channel, is already pushing
-limits and you will have to make tradeoffs in visual quality. 128x256 is definitely pushing
-things and you'll get 100Hz or less depending on the performance options you choose.  
-This puts the maximum reasonable resolution at 384x256 for 3 chains. You can see more examples
-and video capture on [Marc MERLIN's page 'RGB Panels, from 192x80, to 384x192, to 384x256 and maybe not much beyond'](http://marc.merlins.org/perso/arduino/post_2020-03-13_RGB-Panels_-from-192x80_-to-384x192_-to-384x256-and-maybe-not-much-beyond.html)
-
-You can look at this bug for more details: 
-https://github.com/hzeller/rpi-rgb-led-matrix/issues/918#issuecomment-578281400
-
-Basic tips:
+Basic performance tips:
+- Use --led-show-refresh to see the refresh rate while you try parameters
 - use an active-3 board with led-parallel=3
-- led-pwm-dither-bits=1 and led-pwm-bits=7 decrease visual quality but increase refresh frequency
-- You can patch the code to allow led-pwm-lsb-nanoseconds=25 for slightly dimmer output at higher rate
-(some panels can go all the way down to 9 but they get much dimmer)
+- led-pwm-dither-bits=1 gives you a speed boost but less brightness
+- led-pwm-lsb-nanoseconds=50 also gives you a speed boost but less brightness
+- led-pwm-bits=7 or even lower decrease color depth but increases refresh speed
 - AB panels and other panels with that use values of led-multiplexing bigger than 0, 
-will also go faster, although as you tune more options, their advantage will decrease.
+will also go faster, although as you tune more options given above, their advantage will decrease.
+- 32x16 ABC panels are faster than ABCD which are faster than ABCDE, which are faster than 128x64 ABC panels
+(which do use 5 address lines, but over only 3 wires)
+- Use at least an rPi3 (rPi4 is still slightly faster but may need --led-slowdown-gpio=2)
 
-Ultimately, you should not expect to go past 256x256 using an active-3 board without significant
+Maximum resolutions reasonably achievable:
+A general rule of thumb is that running 16K pixels (128x128 or otherwise) on a single chain, 
+is already pushing limits and you will have to make tradeoffs in visual quality. 32K pixels 
+(like 128x256) is definitely pushing things and you'll get 100Hz or less depending on the 
+performance options you choose.  
+This puts the maximum reasonable resolution around 100K pixels (like 384x256) for 3 chains. 
+You can see more examples and video capture of speed on [Marc MERLIN's page 'RGB Panels, from 192x80, to 384x192, to 384x256 and maybe not much beyond'](http://marc.merlins.org/perso/arduino/post_2020-03-13_RGB-Panels_-from-192x80_-to-384x192_-to-384x256-and-maybe-not-much-beyond.html)  
+If your refresh rate is below 300Hz, expect likely black bars when taking cell phone pictures. 
+A real camera with shutter speed lowered accordingly, will get around this.
+
+Ultimately, you should not expect to go past 64K pixels using 3 chains without significant
 quality tradeoffs. If you need bigger displays, you should use multiple boards and synchronize the
 output.
 
--- a/README.md
+++ b/README.md
@@ -688,11 +688,15 @@ appear too dim because each line is not displayed long enough before it is turne
 AB panels also known as outdoor panels, help with performance (you can drive them a bit
 faster and they'll still look bright enough), but as you drive them faster, you'll
 eventually run into the same limits.  
-Ultimately a Raspberri Pi3 is fast enough to push pixels as quickly as panels can accept them.
+Ultimately a Raspberri Pi3 is fast enough to push pixels as quickly as panels can accept them
+(actually, not quite an rPi4 goes a bit faster even with the required --led-slowdown-gpio=2).  
 Using an FPGA solution could work around some slight inefficiencies but ultimately run into 
 similar limits.  
 A general rule of thumb is that running 128x128 on a single channel, is already pushing
-limits and you will have to make tradeoffs in visual quality.
+limits and you will have to make tradeoffs in visual quality. 128x256 is definitely pushing
+things and you'll get 100Hz or less depending on the performance options you choose.  
+This puts the maximum reasonable resolution at 384x256 for 3 chains. You can see more examples
+and video capture on [Marc MERLIN's page 'RGB Panels, from 192x80, to 384x192, to 384x256 and maybe not much beyond'](http://marc.merlins.org/perso/arduino/post_2020-03-13_RGB-Panels_-from-192x80_-to-384x192_-to-384x256-and-maybe-not-much-beyond.html)
 
 You can look at this bug for more details: 
 https://github.com/hzeller/rpi-rgb-led-matrix/issues/918#issuecomment-578281400
--- a/README.md
+++ b/README.md
@@ -678,6 +678,37 @@ the other arguments, no newline). This will use the last core
 only to refresh the display then, but it also means, that no other process can
 utilize it then. Still, I'd typically recommend it.
 
+Performance improvements and limits
+-----------------------------------
+Regardless of which driving hardware you use, ultimately you can only push pixels 
+so fast to a string of panels before you get flickering due to too low a refresh 
+rate (less than 80-100Hz), or before you refresh the panel lines too fast and they
+appear too dim because each line is not displayed long enough before it is turned off.
+
+AB panels also known as outdoor panels, help with performance (you can drive them a bit
+faster and they'll still look bright enough), but as you drive them faster, you'll
+eventually run into the same limits.  
+Ultimately a Raspberri Pi3 is fast enough to push pixels as quickly as panels can accept them.
+Using an FPGA solution could work around some slight inefficiencies but ultimately run into 
+similar limits.  
+A general rule of thumb is that running 128x128 on a single channel, is already pushing
+limits and you will have to make tradeoffs in visual quality.
+
+You can look at this bug for more details: 
+https://github.com/hzeller/rpi-rgb-led-matrix/issues/918#issuecomment-578281400
+
+Basic tips:
+- use an active-3 board with led-parallel=3
+- led-pwm-dither-bits=1 and led-pwm-bits=7 decrease visual quality but increase refresh frequency
+- You can patch the code to allow led-pwm-lsb-nanoseconds=25 for slightly dimmer output at higher rate
+(some panels can go all the way down to 9 but they get much dimmer)
+- AB panels and other panels with that use values of led-multiplexing bigger than 0, 
+will also go faster, although as you tune more options, their advantage will decrease.
+
+Ultimately, you should not expect to go past 256x256 using an active-3 board without significant
+quality tradeoffs. If you need bigger displays, you should use multiple boards and synchronize the
+output.
+
 Limitations
 -----------
 If you are using the Adafruit HAT/Bonnet in the default configuration, then we


marcmerlin:
@hzeller do you feel this would be useful to include so that people have a better idea how far they can push all this now that the panels that let us hit those limits, are readily available?
You can see the rendered version here if needed: https://github.com/marcmerlin/rpi-rgb-led-matrix/blob/patch-3/README.md

hzeller:
The discussion here including the rules of thumb is very specific to your anecdotal experience with 64 pixel high displays (what is an expected maximum refresh etc). However, the range of panels and refresh rates etc. is vastly different and it is hard to suggest what a good expectation should be.

So keep it a bit more generic and not describe what to expect, but to just include tips to describe what parameters can influence the refresh rate and what should be considered

   * One wants to be about 80-100Hz. Ideally 180Hz and more.
   * Ways to achieve that: pwm bits, dither-bits. What are the disadvantages of these (pwm-bits: less colors; dither-bits: less brightness)
   * We are already at the limit of what the panels can do speed-wise, but we're still at the mercy of a non-realtime operating system, so an FPGA solution might not necessarily be faster, but creates less flicker glitches. 
   * Some panels can go fast enough, some need led-gpio-slowdown.
   * other things that might be useful ...

A lot of these things are already described in the [section with the relevant options](https://github.com/hzeller/rpi-rgb-led-matrix#misc-options) so you should emphasize in your paragraph a good systematic approach that goes beyond the flag descriptions how to work on the refresh rate (switch on --led-show-refresh, and which options to tiddle; include led-limit-refresh to counter flicker etc.)

marcmerlin:
thanks for the review. Will rework this accordingly.

marcmerlin:
@hzeller I had another shot.
Hopefully this is closer but if not quite what you wanted still, do you mind merging and touching up accordingly? (will probably quicker than asking me to do it via this issue and going back and forth). 
Don't worry, I won't take offense if you hack it up or remove stuff :)

marcmerlin:
@hzeller gentle ping :)
If it's not exactly what you wanted, are you ok merging and putting your diff/change on top?

marcmerlin:
Hi @hzeller, any time to review this?

hzeller:
Sorry, I am currently involved in a ton of Covid-19 response projects; will get back to this once that load lightens.

marcmerlin:
@hzeller  Just found this abandonned tab in one of my browsers :)  not sure if you're still working on response projects?

hzeller:
I do ( part of https://www.covidshieldnexus.org/ ), but now the 3D printers are mostly humming away on their own.
Let me have a look this weekend.

marcmerlin:
@hzeller  Maybe now is a better time? :(

marcmerlin:
Hi @hzeller  how goes? :)

hzeller:
Cool, thanks!
Sorry for the delay.

arahasya:
Hi,
I am not sure how you have concluded that --led-multiplexing = 0 is the slowest

I am currently in process of manufacturing RGB led boards

I have made 16x48 matrices both with straight scan and interleaving at 48
i.e both --led-multiplexing = 0 , --led-multiplexing = 1
Case A requires 2 RGB data input, Case 2 requires 1 RGB data input
But there is additional track for interleaving at 48 or 32 in case of 16x32
So the tradeoff is double the data on 1 Data line or data on 2 Data lines
2 Data lines seems faster to me

arahasya:
The refresh rate is dependent on scan rows and not on type of multiplexing

daveythacher:
There is two major generations of panel. The second generation can possibly exist in two flavors. To my knowledge this project only supports the first generation panels. The first generation panels use constant current shift registers which have a max clock of 25MHz. This project appears to place a limit several times under that. You may run into cabling issues as the HUB75 technically could have transmission line behavior. Look at IDE/PATA cable history to understand.

However assuming 25MHz you simple take your column times multiplex ratio times PWM step count. So for 32x32 with 1:16 at 256 color steps would have max FPS of 25000000 / (32 * 16 * 256) or 190.7 FPS. Again this assumes the Pi is generating 25MHz and using 256 color steps, which it is not.

It is worth mentioning that likely the panels are reporting row refresh not panel refresh rate. To covert this from the example above multiply by multiplex ratio, 190.7 * 16 or 25000000 / (32 * 256) or 3051.7 Hz.

You can play with these numbers to get better performance in terms of frame rate. Quality will suffer, which is why the second generation panels exist. They were designed to actually handle these problems. The other approach is to increase the number of chains. This project allows up to three. The FPGA controller these are normally paired with use these along with SDRAM/SRAM configurations. The FPGA is not affect in terms of performance the way the Pi is. This is after all IO, which is software's kryptonite.

 Now as you increase the panels you will increase the work load on the CPU. If you increase the bit banging bandwidth you will increase the work load. This will become an issue for stability. Linux will attempt to balance various processes and you can strong arm it to select you more often. However if fully loaded there can be an issue from this alone. There are a number of potential issues that can exist with a software architecture like this. However they not happen.

You can retrofit some of the second generation logic onto the first generation panels with a small FPGA. It will not work as well but will extend the function. You can also use the logic to bridge the Pi to the second generation panels. However this is off topic.

daveythacher:
> The discussion here including the rules of thumb is very specific to your anecdotal experience with 64 pixel high displays (what is an expected maximum refresh etc). However, the range of panels and refresh rates etc. is vastly different and it is hard to suggest what a good expectation should be.
> 
> So keep it a bit more generic and not describe what to expect, but to just include tips to describe what parameters can influence the refresh rate and what should be considered
> 
> * One wants to be about 80-100Hz. Ideally 180Hz and more.
> * Ways to achieve that: pwm bits, dither-bits. What are the disadvantages of these (pwm-bits: less colors; dither-bits: less brightness)
> * We are already at the limit of what the panels can do speed-wise, but we're still at the mercy of a non-realtime operating system, so an FPGA solution might not necessarily be faster, but creates less flicker glitches.
> * Some panels can go fast enough, some need led-gpio-slowdown.
> * other things that might be useful ...
> 
> A lot of these things are already described in the [section with the relevant options](https://github.com/hzeller/rpi-rgb-led-matrix#misc-options) so you should emphasize in your paragraph a good systematic approach that goes beyond the flag descriptions how to work on the refresh rate (switch on --led-show-refresh, and which options to tiddle; include led-limit-refresh to counter flicker etc.)

The FPGA can handle the full display. However supporting the array of panels can take some work. Simple hard coded FPGAs will be fast, but flexible ones can be expensive and larger setup. There is a way to make a flexible setup for low cost however it does not remove the variations in Linux/software on Pi.

Dithering and PWM reduction simply reduce CPU and bandwidth overhead. It moves the limit down, but would work to some degree. Not for larger panels and high refresh panels.

daveythacher:
> @hzeller do you feel this would be useful to include so that people have a better idea how far they can push all this now that the panels that let us hit those limits, are readily available?
> You can see the rendered version here if needed: https://github.com/marcmerlin/rpi-rgb-led-matrix/blob/patch-3/README.md

The best thing to do is track down how this project sets the HUB 75 clock. List a few example speeds, with noting this is dependent on idle system. There may be some difference between the Pi models, but this project may compensate for that. ( I do not recall.) Then give overview of PWM bit, multiplex ratio and column depth affect on overall refresh rate. You should be able to estimate this at least for idle system.

Note the PWM bits is not exactly the traditional PWM bit count. In standard terms with 24 bit color you have 16.7 million unique colors. Things would a little different with this project since it works off dither. What matters for refresh is what moves across the HUB 75 wires.

The max performance is set in terms of bandwidth, There is only enough GPIO on the Pi for three channels assuming you do not need pins for other things. The clock is limited, therefore to improve the refresh or allow larger panels you have to change the image quality settings.

I attempted to explain an example above. The documentation as I recall mentioned it was good enough for 3 chains of 5 32x32 panels. This could put the clock at something like 1 or 2 MHz. However I did not look very far and it will need to be verified. Perhaps Hzeller knows.

daveythacher:
This is worth a link: https://en.wikipedia.org/wiki/List_of_monochrome_and_RGB_color_formats#9-bit_RGB

hzeller:
The speed is set with the gpio slowdown, and with a current Pi4, we need to slow down things a lot to work with the panels to be in the 20Mhz range.

The color output is of course not done with PWM (which would be strongly limited by the clock speed) but binary code modulation 11 bit resolution. The clock speed of the panels still affect the few lower bits as limiting factor. The 11 bit linear resolution is needed to get enough detail to satisfactorily convert a 8-bit value with exponential luminance curve. So from a marketing perspective, we could say 33 bit color resolution ... but of course that translates to regular 24 bit color.

There are many parameters that influence each other (e.g. the length of the chain determines how much we're waiting for the clocking vs. the Output Enable. As we parallel clock things in, we're only limited by the clocking in the lower bits) that I usually suggest to do a dry-run on a Pi with `--led-show-refresh` is someone wants to determine the refresh rate.

daveythacher:
> The speed is set with the gpio slowdown, and with a current Pi4, we need to slow down things a lot to work with the panels to be in the 20Mhz range.
> 
> The color output is of course not done with PWM (which would be strongly limited by the clock speed) but binary code modulation 11 bit resolution. The clock speed of the panels still affect the few lower bits as limiting factor. The 11 bit linear resolution is needed to get enough detail to satisfactorily convert a 8-bit value with exponential luminance curve. So from a marketing perspective, we could say 33 bit color resolution ... but of course that translates to regular 24 bit color.
> 
> There are many parameters that influence each other (e.g. the length of the chain determines how much we're waiting for the clocking vs. the Output Enable. As we parallel clock things in, we're only limited by the clocking in the lower bits) that I usually suggest to do a dry-run on a Pi with `--led-show-refresh` is someone wants to determine the refresh rate.

Maybe I am dense. However I fail to see how there is not some number of PWM steps moving across the wire. These are shift registers so you have to indirectly toggle them. If any one of the 6 data pins has to change then you need to send it over for all.

I get how the image generation side can use algorithms to trick the human eye or create low bit RGB color schemes. However the image quality is less, it is like comparing interlaced to progressive scan.

In any case the real question is what is the effective step number and clock. If 20 MHz then you are saying your example would be blasting 710 FPS for 160x32 in a single chain. Is this correct? 20,0000,000/(160*16*11)

Also my point is your effective color palette may be something like 11^3 or somewhere between 9 and 12 bit RGB color. I am ignoring the OE time, which can have a decent effect depending on things.

hzeller:
Yes, the 710fps figure is right if you dont' take OE time into account. But OE is the dominant part of the time spent for the highter bits.
So in practice you'd get maybe 400Hz if you choose low 50ns LSB OE and 11 bit.
If you choose to dither the lower bits you might get 600-800 fps with that setting.

Clock rate needs to be tweaked with the panels at hand, some are faster, some are slower; also how short the cables are etc.

daveythacher:
Does the setting "--led-pwm-lsb-nanoseconds" change the time for both the clock and OE? OE apples at row change time, does each row get shifted out 11 times or are all rows scan 11 times?

You store in 11 bits of information, but if I am not mistaken you shift. So you do not use the complete number range. You use 11 bits to hold 11 values. Therefore you have 11 values not 2048 values of unique color.

hzeller:
No, the`--led-pwm-lsb-nanoseconds` just influences the on-time for the LSB.

The 11 bits are bitplanes. Each bit plane is shifted out and left on with output enable a different time: `(lsb-nanoseconds << plane#)`. So the lowest bit-plane stays on for, say, 100ns, the next for 200ns, then 400ns and so on...
That way, once we have done this for all bits, the sum of on-time represents the value 0..2047.
This is commonly referred to as binary code modulation.

daveythacher:
> No, the`--led-pwm-lsb-nanoseconds` just influences the on-time for the LSB.
> 
> The 11 bits are bitplanes. Each bit plane is shifted out and left on with output enable a different time: `(lsb-nanoseconds << plane#)`. So the lowest bit-plane stays on for, say, 100ns, the next for 200ns, then 400ns and so on...
> That way, once we have done this for all bits, the sum of on-time represents the value 0..2047.
> This is commonly referred to as binary code modulation.

That is interesting. I think I understand what you are doing now. However there is a small problem here. This is an optimization for hardware. The fundamental bandwidth is the same, which is governed by the lowest bit. (I think on this part we agree.) 

However you are putting even more pressure than I realized on real time control. This does save a ton of cycles though, but I am not really sure that matters. Since you can generate the image and store in memory then play it back. I mean the number of cycles alone required to make up for that bandwidth especially with bit banging. It will still be less but in this case it ends up wasting a lot. This is common problem for bit banging and super loop logic.

This approach is more efficient for hardware generating PWM locally and reduces the number of cycles moving across the HUB75. However the bandwidth is fundamentally the same. I also understand the need for refresh improving approaches of E-PWM/S-PWM if this is really what they do in PWM.

Here is a link which glosses over the issue of low ns real time pressure: https://hackaday.com/2011/07/22/using-binary-code-modulation-to-control-led-brightness/

I looked into those questions myself. That setting only defines the min unit of bandwidth. (As the name implies.) Note this is BCM no PWM! You also do scan all rows before scanning to next bit. Which you could avoid and help improve the effect of the OE. However since you support large bit values in monolithic approach this is probably simpler.

BCM will help with OE delay, but will cause issues with large bit values. Heck I question if it is supported still. It just seems odd to me 2048 * 50ns against 1/20000000 seconds times column and multiplex rate. How the heck there is time, which is why I guess you have to reduce it.

In summary the formula becomes: 1/(PWM_LSB_NS * (columns + 1) * multiplex ratio * BCM bits)

Sound about right?

hzeller:
* The lowest bits indeed take longer to clock than the on-time which follows, so there is a bit of dark-time for these bits. However for the higher values, the next row can be clocked in completely while the on-time is still ongoing.
 * In lower bits, the clocking in data dominates the time, in higher bits, OE dominates the time. So for a refined calculation, you've to determine for each bit-plane value `max(PWM_LSB_NS * 2^bit, serial_clock_cycle * columns)` to determine how long each bit takes. Then multiply that with the multiplex ratio.
 * Not sure where you see the rows are scanned first; each row goes through the full BCM cycle before the next row is tackled (it would also not work, as the LED matrix have quite some transition time between rows, visible as significant flicker)
 * The realtime requirements are not as much as the OE timing is actually done by hardware (I am using a circuit in the BCM that directly outputs to a pin) so indepenend of anything Linux can mess with. With increased pressure close to 100% CPU though, this still can interrupt visibly (which is why you use limit refresh then).
 * With naive PWM you would not even get near any useful refresh rate.

daveythacher:
Conceptually PWM in software is a serious pain due to memory and/or processor cycle usage. Conceptually BCM is a pain in software for real time/interrupt response. PWM in hardware is huge and power hungry compared to BCM. Hardware can fix the both version's software issues. 

There is a reason to use PWM in software to a degree, however not a good one. PWM for LEDs creates its own frequency like BCM does. Which is called the frame rate. There is a maximum bandwidth here which creates this frequency. However you can use that bandwidth as a clock source for PWM to ensure correct timing conceptually. Both PWM and BCM need a counter this is true for hardware and software.

The difference is how it is formed. PWM uses it directly, while BCM responds to it. Here the nature of the processor becomes important. Hence the real time issue, especially for something that cannot ensure high frequency events beyond a point. However if one is able to accelerate the computation of values it can be automated with hardware. 

SPI, DMA, NEON, FPGA, etc. all provide various offloads which allow the bulk of the work to be done in a non real time system, so long as the system can feed data within time spec. Full offload and real time control would likely involve a hardware process which has access to memory. Here architecture issues of subsystem can matter.

Now in the context of bit banging and super loops there is no real insurance outside the clock driving the software itself. Which means you either slow/waste cycles or switch to interrupts. The subsystem for interrupts then matters. This can have a serious impact on BCM and PWM.

PWM can be viewed as 32 times slower than BCM for 256 steps in software. Since it only handles the 8 bits versus 256 bits per LED. Now the interesting thing is this also means that the HUB75 clock could be viewed as reduced by 32. However you still have to support the lower bits which are high speed, so you still have the original bandwidth.

BCM has better memory expansion footprint and lower cycle for runtime generation of values. However due to the bandwidth you still end up having to allocate a huge amount of cycles. Unless you plan to program it in assembly or tightly couple everything. Which will create problems for other things. This is true for software at least.

The time when this bandwidth frequency does not exist is when you are driving the LEDs directly versus indirectly via shift registers for example. I will concede the BCM is better than PWM still. However I do not believe it fundamentally affects most of my points. Algorithmically better due to direct nature of expression, however the direct nature fundamentally still remains unchanged to the degree I was working under.

daveythacher:
> * The lowest bits indeed take longer to clock than the on-time which follows, so there is a bit of dark-time for these bits. However for the higher values, the next row can be clocked in completely while the on-time is still ongoing.
> * In lower bits, the clocking in data dominates the time, in higher bits, OE dominates the time. So for a refined calculation, you've to determine for each bit-plane value `max(PWM_LSB_NS * 2^bit, serial_clock_cycle * columns)` to determine how long each bit takes. Sum that up for all bits. Then multiply that with the multiplex ratio.
> * Not sure where you see the rows are scanned first; each row goes through the full BCM cycle before the next row is tackled (it would also not work, as the LED matrix have quite some transition time between rows, visible as significant flicker)
> * The realtime requirements are not as much as the OE timing is actually done by hardware (I am using a circuit in the BCM that directly outputs to a pin) so indepenend of anything Linux can mess with. With increased pressure close to 100% CPU though, this still can interrupt visibly (which is why you use limit refresh then).
> * With naive PWM you would not even get near any useful refresh rate.

Never mind, I think you are correct on the bit planes. My equation is not correct, well I tried. However I think you may have glitch.

daveythacher:
> * With naive PWM you would not even get near any useful refresh rate.

Speaking of "naive PWM" there is a possible issue with BCM at signal level. I will just leave it like this.

PR #980, opened 2020-02-04
Implement z-arrangement pixel mapper

lehni:
This pixel mapper allows for these kind of arrangements:
```
1 … 2
  /
3 … 4 
  /
5 … 6
```

### DIFF
--- a/lib/pixel-mapper.cc
+++ b/lib/pixel-mapper.cc
@@ -217,7 +217,49 @@ private:
   int parallel_;
 };
 
+class ZArrangementMapper : public PixelMapper {
+public:
+  ZArrangementMapper(): chain_(0), panel_columns_(2) {}
 
+  virtual const char *GetName() const { return "Z-mapper"; }
+
+  virtual bool SetParameters(int chain, int parallel, const char *param) {
+    chain_ = chain;
+    if (param != NULL) {
+      char *errpos;
+      panel_columns_ = strtol(param, &errpos, 10);
+      if (*errpos != '\0') {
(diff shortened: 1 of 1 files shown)


hzeller:
This doesn't look like it takes parallal chains into account in which the parallel chains are also arranged that way. Can you update the code to do that ?
(This would be useful for #1003 )

marcmerlin:
@lehni  If you have a parallel output board, can you look at Vmapper that just got submitted and would allow you to run 1-3-5 and 2-4-6 in parallel?
Also, if you have a single chain and turn your panels 3 and 4 upside down to have the shortest cables possible, wouldn't that work with the existing U-mapper?

lehni:
I don't have a parallel board. Maybe somebody else can take it from here? I wrote this mostly for my own needs and thought I could share it, but I don't have the time to work more on it for a feature that I don't need and can't test. I hope you understand.

marcmerlin:
@lehni , Henner just made me realize that my code already does that you needed, so we don't need to merge yours anymore.
Let's assume that your panels are 64x32:
please try mapper=V-mapper --led-rows=32 --led-rows=128 --chain=3 --mapper=V-mapper
This should just work with your panels by pretending that your 2 panels wide are a single panel of 128 wide.

Can you confirm?

lehni:
Sadly I don't have access to the setup right now due to the Corona / WFH rules in Switzerland. I don't know when I'll next be in a situation where I can test this, but I will let you know when I did.

hzeller:
Could you confirm if @marcmerlin 's V-Mapper helps your use-case @lehni ?

PR #981, opened 2020-02-09
Update framebuffer.cc adding SM5266 row setter and FM6127 initialization

gohlhausen:
Since this adds a new row decoder, the decoder checking part also needs to change in options-initialize.cc

  if (row_address_type < 0 || row_address_type > 4) {
    err->append("Row address type values can be 0 (default), 1 (AB addressing), 2 (direct row select), 3 ABC address.\n");
    success = false;
  }

### DIFF
--- a/lib/framebuffer.cc
+++ b/lib/framebuffer.cc
@@ -112,6 +112,56 @@ private:
   int last_row_;
 };
 
+class SM5266RowAddressSetter : public RowAddressSetter {
+public:
+  SM5266RowAddressSetter(int double_rows, const HardwareMapping &h)
+    : row_mask_(0), last_row_(-1),
+      bk_(h.c),
+      din_(h.b),
+      dck_(h.a)
+  {
+
+    assert(double_rows <= 32);  // need to resize row_lookup_
+    if (double_rows > 16) row_mask_ |= h.e;
+    if (double_rows > 8)  row_mask_ |= h.d;
+    for (int i = 0; i < double_rows; ++i) {
+      // To avoid the bit-fiddle in the critical path, utilize
+      // a lookup-table for all possible rows.
+      gpio_bits_t row_address = 0;
+      row_address |= (i & 0x08) ? h.d : 0;
+      row_address |= (i & 0x10) ? h.e : 0;
+      row_lookup_[i] = row_address;
+    }
+  }
+
+  virtual gpio_bits_t need_bits() const { return row_mask_; }
+
+  virtual void SetRowAddress(GPIO *io, int row) {
+    if (row == last_row_) return;
+    io->SetBits(bk_);
+    for (int r = 0; r < 8; r++) {
+      if (row % 8 == (7-r)){
+        io->SetBits(din_);}
+      else{
+        io->ClearBits(din_);}
+      io->SetBits(dck_);
+      usleep(1);
+      io->ClearBits(dck_);
+      }
+    io->ClearBits(bk_);
+    last_row_ = row;
+    io->WriteMaskedBits(row_lookup_[row], row_mask_);
+  }
+
+private:
+  gpio_bits_t row_mask_;
+  int last_row_;
+  const gpio_bits_t bk_;
+  const gpio_bits_t din_;
+  const gpio_bits_t dck_;
+  gpio_bits_t row_lookup_[32];
+};
+
 class ShiftRegisterRowAddressSetter : public RowAddressSetter {
 public:
   ShiftRegisterRowAddressSetter(int double_rows, const HardwareMapping &h)
@@ -365,6 +415,9 @@ Framebuffer::~Framebuffer() {
   case 3:
     row_setter_ = new ABCShiftRegisterRowAddressSetter(double_rows, h);
     break;
+  case 4:
+    row_setter_ = new SM5266RowAddressSetter(double_rows, h);
+    break;
   default:
     assert(0);  // unexpected type.
   }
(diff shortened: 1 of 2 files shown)


lib/framebuffer.cc
```diff
@@ -112,6 +112,56 @@ class DirectRowAddressSetter : public RowAddressSetter {
...
+    }
+  }
+
+  virtual gpio_bits_t need_bits() const { return row_mask_; }
+
+  virtual void SetRowAddress(GPIO *io, int row) {
+    if (row == last_row_) return;
+    io->SetBits(bk_);
+    for (int r = 0; r < 8; r++) {
+      if (row % 8 == (7-r)){
```
hzeller: Can you make the braces similar to the rest of the code ?
```
  if (foo) {  // <- space before {
    doSomething();
 } else {     // <- closing curly brace on separate line
    doSomethingElse();
 }

File: lib/framebuffer.cc [hunk 2]
```diff
@@ -112,6 +112,56 @@ class DirectRowAddressSetter : public RowAddressSetter {
...
 
+class SM5266RowAddressSetter : public RowAddressSetter {
+public:
+  SM5266RowAddressSetter(int double_rows, const HardwareMapping &h)
+    : row_mask_(0), last_row_(-1),
+      bk_(h.c),
+      din_(h.b),
+      dck_(h.a)
+  {
+
```
hzeller: Please don't leave empty lines.

lib/framebuffer.cc
```diff
@@ -427,13 +480,126 @@ static void InitFM6126(GPIO *io, const struct HardwareMapping &h, int columns) {
...
+    if (i > columns - 11) value |= h.strobe;
+    io->Write(value);
+    io->SetBits(h.clock);
+    io->ClearBits(h.clock);
+  }
+  io->ClearBits(h.strobe);
+
+}
+
+
```
hzeller: Just leave one empty line here.

lib/framebuffer.cc
```diff
@@ -112,6 +112,56 @@ class DirectRowAddressSetter : public RowAddressSetter {
   int last_row_;
 };
 
+class SM5266RowAddressSetter : public RowAddressSetter {
```
hzeller: Could you write a brief description of how the setting of bits works here ?
Looks like the lower 3 bits are clocked in to din_, while bit 4 and 5 are set directly ?

(it sounds weird that they do this, while they have perfectly good three direct bits available that they now use to clock in exactly three bits; possibly a result for how they internally do the multiplexing ?)

lib/framebuffer.cc
```diff
@@ -112,6 +112,56 @@ class DirectRowAddressSetter : public RowAddressSetter {
...
+public:
+  SM5266RowAddressSetter(int double_rows, const HardwareMapping &h)
+    : row_mask_(0), last_row_(-1),
```
hzeller: You want to set `row_mask_` to `h.a | h.b | h.c`, otherwise these bits are not initialized for output.

lib/framebuffer.cc
```diff
@@ -112,6 +112,62 @@ class DirectRowAddressSetter : public RowAddressSetter {
...
+// The panel this works with has 8 SM5266 shifters (4 for the top 32 rows and 4 for the bottom 32 rows).
+// DE is used to select the active shifter (rows 1-8/33-40, 9-16/41-48, 17-24/49-56, 25-32/57-64).
+// Rows are enabled by shifting in 8 bits (high bit first) with a high bit enabling that row.
+// This allows up to 8 rows per group to be active at the same time (if they have the same content), but that isn't implemented here.
+// BK, DIN and DCK are the designations on the SM5266P datasheet.  
+// BK = Enable Input, DIN = Serial In, DCK = Clock
+class SM5266RowAddressSetter : public RowAddressSetter {
+public:
+  SM5266RowAddressSetter(int double_rows, const HardwareMapping &h)
+    : row_mask_(h.a | h.b | h.c | h.d | h.e),
```
hzeller: Here, you only need h.a, h.b and h.c. The h.d and h.e we decide later if we need them in line 132.

Reason why we do this is that we then only claim the GPIO ports we need to control the library, while the remaining pins stay free for the user to use for something else.

I'll fix it.

marcmerlin:
Hi @gohlhausen  and @hzeller  I just received some newer panels which turned out had FM6127 instead of FM6126A. They almost work, except for chaining:
image

If I put the panel first in the chain, it's all good, but by the time it's 3rd, I pseudo randomly get some sections that do not get initalized correctly and a color component won't turn on.
If I go back to FM6126A with the same rPi4 and cables, it works perfectly, so I think the tming for FM6127A may be off by just a bit?

I checked my tree, and it was running
commit c6c2076d4444ba70a93a55a627b63155ccf54277
Author: Henner Zeller <h.zeller@acm.org>
Date:   Mon Mar 16 18:38:21 2020 -0700

which is newer than
commit 144dd74fd48b713e860ac88a09c52eab9a45513f
Author: Henner Zeller <h.zeller@acm.org>
Date:   Mon Feb 10 22:53:57 2020 -0800
    Update README with FM6127 reference.
commit 1fd2751c212c5ec8ba5f0146225e7550ca78f6ea
Author: gohlhausen <gaylano@gmail.com>
Date:   Tue Feb 11 00:27:56 2020 -0600
    Update framebuffer.cc adding SM5266 row setter and FM6127 initialization (#981)

marcmerlin:
this belongs in a bug to be easier to reference, so I just made one: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1103

PR #984, opened 2020-02-13
Added set image function

wolfi-by:
The goal of this change is to add a function to the CANVAS that can transfer an Color-Array of RGB or BGR values into the library. The color byte order is selected by parameter. The Function is called by using SetImage(IntPtr *byte, int length, bool isBGR) where the IntPtr points to an byte array that contains all pixel data, length is the length of the given array in bytes and at least the bool value to set the byte order inside the array.
Additionally inside that function it is tested, if the Colorvalue is Black. When black is detected nothing is done and the next pixel will be processed. 
The goal of this function is to raise speed of drawing images when not the whole canvas is painted with color. Especially in my case it was useful because of many black areas without a change to know on which position color data is set. It is quite useful in combination using a network stream or a dotnet Bitmap (System.drawing.Bitmap). 

I made some tests by declaring a pixel matrix with following parameters:
64x64 pixel matrix,
3 parallel,
10 per row

I wrote a Text ("empty" onto the Matrix using Verdana.ttf (Windows font) with height of 20 pixel on a random position. Background of the matrix was black, Fontcolor red including Anti Alias.
I used a RaspberryPi4 with Raspbian lite, this library for drawing, dotnet core for processing and creating image. Controlled via Blazor Webpage.
With that setup i got following timing results:
Creating the Bitmap takes about 1.3 Milliseconds.
Drawing the pixels onto the Matrix by using SetPixel took about 150ms.
Drawing by using the Function SetImage about 5ms.

Greetings.
PS.: I am not good in programming c++, maybe it is good to review my code. in my setup it ran quite good and i hope it is helpful for someone.


wolfi-by:
Compare schoss changees on 140? Files... I think i only changed 5

hzeller:
It looks like the file line-endings were changed from linefeed to carriage-return-line-feed. Did you happen to edit this on a Windows machine ? This probably changed all the files.

Can you configure your set-up to keep the line-endings ?

KentWalker:
If using Windows OS I suggest using Notepad++ to edit your files (or parse after editing), you have the option of setting EOL terminators to either Windows, Mac or Unix.

hzeller:
@wolfi-by do you have an update with the line-endings corrected ?

wolfi-by:
Ah sorry... I havent seen your comment... I will Do until tomorrow.

wolfi-by:
I changed all the files from \r\n to \n as mentioned above... Seems that there are less differences now. But I dont know how to work with the confilcts. I think its better for me to ask a friend how to deal with so I think on Monday the work is done.

lib/transformer.cc
```diff
@@ -81,91 +88,130 @@ void RotateTransformer::TransformCanvas::SetPixel(int x, int y, uint8_t red, uin
   }
 }
 
-int RotateTransformer::TransformCanvas::width() const {
+void RotateTransformer::TransformCanvas::SetImage(uint8_t *rawdata, int length, bool isBGR)
```
hzeller: Make sure to use the curly brace rules used in the rest of the file (open curly braces after parenthesis etc).

This could probably be formulated a bit more efficiently to reduce calculations

   * store width and height in a local variable beforehand so that we don't have to call a virtual method in each loop iteration.
   * it might be slightly faster to swap the loops and have outer loop y and inner loop x, and in the inner loop rawdata += 3 after setting the pixel. Thus, we don't have to calculate the offset every time
   * if you are concenred about performance, maybe putting the `if (isBGR) ...` outside and have two slightly different versions of the loop could avoid one branch in the inner loop. Might be worthwhile to benchmark, it might not be too much of a difference.

hzeller: This implementation should go into the base Canvas class: it is always the same for everyone.
It can still be virtual, so that  special implementations might be able to override it if they have a faster implementation for their use-case.

hzeller:
Since I have the file open in my editor, let me prepare the SetImage() function in the Canvas then you can wire up the functionality in C#.

hzeller:
In `graphics.h`, I've now added a free function [`SetImage()`](https://github.com/hzeller/rpi-rgb-led-matrix/blob/cf3c3c2ef0b1d1a637c1b9590965186a05142f5c/include/graphics.h#L91) including the `is_bgr` flag in commit
https://github.com/hzeller/rpi-rgb-led-matrix/commit/cf3c3c2ef0b1d1a637c1b9590965186a05142f5c

There is also the corresponding [`set_image()`](https://github.com/hzeller/rpi-rgb-led-matrix/blob/cf3c3c2ef0b1d1a637c1b9590965186a05142f5c/include/led-matrix-c.h#L271) in the C-API.

I decided to make it a free function instead of a member function of `Canvas` as it is not something that needs the internals of Canvas and also has some very specific signature; to keep the implementation decision burden low for implementors of `Canvas`, this is good.

Now for your project, you just need to adapt the c# binding to also provide access to this set image function.

hzeller:
I've made the [`SetImage()`](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/include/graphics.h#L83-L105) function a little more universal now so that you can set any size image and also can specify an offset. This then makes scrolling easy.

wolfi-by:
ok... i will take a look at changes... sounds great!

Issue #1003, opened 2020-03-14
No support for panels stacked up high instead of wide?


marcmerlin:
With 12 panels, with 3 chains of 4 panels of 128x64, I need the chains to be
128x256 (4 panels of 128x64 on top of one another)
I'm not sure the mapper function allows this, and if I go 4 panels wide, I get 512x64 x 3 = 512x 192
when I'd much rather have 128x256x3 = 384x256.
Is that possible at all with existing --led-pixel-mapper options ? (it didn't seem like it)
If so, am I stuck having to write my own mapper as per 
https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/examples-api-use#remapping-coordinates ?

marcmerlin:
![IMG_20200313_191048](https://user-images.githubusercontent.com/1369412/76673063-26793e80-655f-11ea-8117-dc4c40a1f271.jpg)

hzeller:
Yes, writing a pixel-mapper would be the way to go
https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/examples-api-use#writing-your-own-mappers

Ideally, we had a generic mapper that parses some configuration string describing the layout; something like a newline separated format describing the physical layout with each line with panels represented as square-bracket enclosed `<parallel-pos>:<chain-pos>` text in it, possibly with a modifier (one of `<`, `^`, `v`, '>') to describe in which direction the data cable is plugged in; so your layout could be something like
```
[0:0<][1:0<][2:0<]
[0:1<][1:1<][2:1<]
[0:2<][1:2<][2:2<]
[0:3<]]1:3<][2:3<]
```

I've not implemented such mapper yet, I'd like to leave that as an exercise to whoever needs it first :)

marcmerlin:
@hzeller thanks for confirming, and smart man about leaving this as a programming exercise to the reader :)
I'll see if I can con myself in yet another rabbithole / yack shaving (I have enough for a few sweaters by now). If you are curious, this all started with:
- a 24x32 neomatrix array (neopixels). To get more speed, I had to port Adafruit::NeoMatrix to FastLED, and wrote FastLED::NeoMatrix
- needed "a few" more pixels, so I had to switch to SmartMatrix, but the API was all different, so I had to write SmartMatrix::GFX to put the same API on top of SmartMatrix
- I'll save you what I learned and had to work around on, regarding ESP32 memory management
[...]

PR #1009, opened 2020-03-18
Vmapper

marcmerlin:
Fixes https://github.com/hzeller/rpi-rgb-led-matrix/issues/1003

Ok, this was more work than I thought for something seemingly simple (well, it's likely just that I suck at it), but it works!
Hopefully this puts me in the lead for the highest usable resolution display on a single rPi  (384x256) :)
401_20200310_FastLED_RPIRGBPanel_GFX_384x256


hzeller:
Cool!
Only potential user-support issue I see is that the panels are fed from the top while the 'usual' way is to feed them from the right.

So to reduce potential user-support in the future, I suggest to not deviate too much from the 'things are fed from the right side', and think about stuff rotated by 90 degrees (it is probably just a little x/y swap fiddle somewhere).

To make sure there is no confusion about how to connect each of these panels, can you add an text-art that indicates how they are exactly chained ? I find that the Macron-Character is excellent for this kind of stuff:
```
[v]¯\_[v]¯\_[v]¯\_[v]¯ ch1
[v]¯\_[v]¯\_[v]¯\_[v]¯ ch2
[v]¯\_[v]¯\_[v]¯\_[v]¯ ch3
```

(if you still want to operate it from the top in your specific application, you then can use the Rotate transformation to get it back rotated)

hzeller:
(also, this brings out much more of the `V`-ness)

marcmerlin:
Mmmh, so the whole point of vmapper is that the panels are fed up-down. Sure, I can rotate everything and then force the use of a second rotate:90 mapper to get it the way you're suggesting, but the whole point of V-mapper is that the panels are fed from up to down (or down to up).

I think you're saying that I can pretend that my panel is like this, is that what you meant?
image

If so, I'm not sure I'm not sure if your suggestion actually makes thing more obvious given that the point of V stacking is that you want matrix is naturally higher than wide, so it's not very obvious to rotate it 90 degrees.
That said, if somehow you have a usage case where you want it that way, you can always use rotate:90, but by default I think it ought to be the way I coded it, wouldn't you agree?
I'm happy to change the ASCII to show how the wiring is (as per your suggestion).
Would that sole change be ok?

hzeller:
mmh ok, so the reason why you want to feed it from the top is that then the panels are 'upright' as they are meant to be ? (because sometimes the LED shielding sideways is a little different than up/down). I could be convinced by that.

Let's do the ASCII art improvement for now, but since you turn it 90 degrees than in my suggestion, you probably need some sort of 'right side pipe symbol' and 'left side pipe symbol' and backslash to illustrate.

marcmerlin:
Just to answer your question, yes, the panels are upright as they were meant to be.
This is their natural position given the writing on the back, so I felt it made sense to keep them that way by default.
Of course if someone needs a resolution that is the other way around, they can use rotate.

examples-api-use/README.md
```diff
@@ -313,12 +313,15 @@ what you can reasonably output without having an unusable fresh rate
...
-  Ch1 Ch2 Ch3
-   v   v   v
-  [>] [>] [>]
-  [>] [>] [>]
-  [>] [>] [>]
-  [>] [>] [>]
+  Ch1     Ch2     Ch3
+   v       v       v
+  [I > O] [I > O] [I > O]
+   ,---'   ,---'   ,---'
```
hzeller: Oh, pretty :)

lib/pixel-mapper.cc
```diff
@@ -217,6 +217,54 @@ class UArrangementMapper : public PixelMapper {
...
+
+
+class VerticalMapper : public PixelMapper {
+public:
+  VerticalMapper() {}
+
+  virtual const char *GetName() const { return "V-mapper"; }
+
+  virtual bool SetParameters(int chain, int parallel, const char *param) {
+    if (chain < 2) {
```
hzeller: Why this limitation ? With a chain of one it would be possible to make a long panel (e.g. for a runtext) that is fast to update (because it then would be three short segments, that make use of parallel feeding)

```
[O < I] [O < I] [O < I]
     |       |       |
     ch3     ch2      ch1
```

Looking at the code below, it _should_ be not a problem (we never divide by (chain_ - 1) for instance).

marcmerlin:
@hzeller your are absolutely correct that chains of 1 would both work, and make sense. I updated that.
You are also correct that I got the ASCII art reverted (not counting left-right which is a bit tricky as it depends on whether you are showing the panels with pixels facing you or with the connectors facing you, but I think that one isn't too important, the up/down one is; that said I changed both).

lib/pixel-mapper.cc
marcmerlin: @hzeller  your are absolutely correct that chains of 1 would both work, and make sense. I updated that.
You are also correct that I got the ASCII art reverted (not counting left-right which is a bit tricky as it depends on whether you are showing the panels with pixels facing you or with the connectors facing you, but I think that one isn't too important, the up/down one is; that said I changed both).

hzeller:
Do you have a picture of the panel backside ? Every panel I have ever seen has the input on the right side (seeing from the face of the panel), not on the left.

hzeller:
Looking at the three-chained V-example you have above, I don't see arrows but the text is indeed up only if the input is on the right.

So looking to the front of the panel, the input is on the right, so we need to make sure to have the code reflect that.

marcmerlin:
IMG_20200318_090455

hzeller:
See, input is on the right of the panel, output on the left
(looking from the back, the input is on the left, output on the right)

marcmerlin:
connector is left when seen from the back, right when seen from the LED (front) side.
The new graph I sent you is showing I on the left which is where it is when you are wiring the panels and the connectors are facing you.
Regardless of whether we show one side or the other, it should be clear enough to someone repeating this hopefully :)

hzeller:
The wiring diagrams in all of the rpi-rgb-led-matrix done are all with the panel fronts looking toward you.
So if you have this already wired and programmed this way, then just the graphics needs to be flipped.

marcmerlin:
Sure, PTAL

examples-api-use/README.md
```diff
@@ -291,6 +291,40 @@ two chains with 8 panels each
...
+```
+  [O < I] [O < I] [O < I]
+   ,---'   ,---'   ,---'
+  [O < I] [O < I] [O < I]
+   ,---'   ,---'   ,---'
+  [O < I] [O < I] [O < I]
+   ,---'   ,---'   ,---'
+  [O < I] [O < I] [O < I]
+       ^       ^       ^
+      Ch1     Ch2     Ch3
```
hzeller: After flipping, wouldn't now the ch1 be on the right ?

marcmerlin: you are correct, sorry about that, fixed.

examples-api-use/README.md
```diff
@@ -291,6 +291,40 @@ two chains with 8 panels each
...
+mapper, you can either do a matrix of 3x3 panels (384x192). Adding more panels
+would force you to have a 4x3 layout (512x192) while you probably want a 3x4
+layout (384 * 256) for a better aspect ratio.
+
+```
+  ./demo --led-rows=64 --led-cols=128 --led-chain=4 -led-parallel=3 --led-pixel-mapper=V-mapper -D0
+```
+
+Note that this is not a U mapper, all the panels are correct side up, and you need
+more cable length. It looks like this, 3 chains of 4 panels for a total of 32K pixels
```
hzeller: 12 panels with 128x64 pixels each is actually a almost 100k pixels (98304)...

examples-api-use/README.md
```diff
@@ -291,6 +291,40 @@ two chains with 8 panels each
...
+per chain. This is also a good time to mention that this is probably an upper limit of
```
hzeller: ... ah, just see the 'per chain :)' nvm.

lib/pixel-mapper.cc
```diff
@@ -217,6 +217,50 @@ class UArrangementMapper : public PixelMapper {
...
+
+  virtual void MapVisibleToMatrix(int matrix_width, int matrix_height,
+                                  int x, int y,
+                                  int *matrix_x, int *matrix_y) const {
+    int panel_width  = matrix_width  / chain_;
+    int panel_height = matrix_height / parallel_;
+    *matrix_x = (x % panel_width) +  int(y/panel_height)* panel_width;
+    *matrix_y = (y % panel_height) + int(x/panel_width) * panel_height;
+
+    //fprintf(stderr, "%s: Panel-W:%d Panel-H:%d. X: %3d -> %3d, Y: %3d -> %3d\n",
```
hzeller: Don't leave commented-out code in the final version.

hzeller:
Cool, almost ready to go: just some leftover debug printf to be removed.

lib/pixel-mapper.cc
marcmerlin: So, I left it there on purpose it case it helps someone (possibly me) writing the next mapper. Understanding how things worked without it would have been hard (it was non obvious to see how things were mapped)
If you'd prefer, I can put it under an #ifdef 0  or something else, though, but if you really don't want it, I can remove it and apply it again in my tree as I will definitely want it again in the future if I write another mapper.
What's your call?

hzeller: Removing commented-out here. Someone who will want to debug this will add something very similar in the working copy, but here, we should leave out the clutter.

Also above in the GetSizeMapping(), the output should not be there, but you can `#if 0` that out if you want.

marcmerlin: Ok, PTAL

hzeller:
Thanks for your contribution! Merged.

hzeller:
Do you have a picture of the back of your panel that you'd like to share ? I think similar to the U-mapper example image, that will help people in the wiring.

marcmerlin:
@hzeller I don't have rails, it's sadly all like just sitting on my desk, so I can't really flip it around
The picture I gave upthread of a single set of 3 panels is basically what a chain looks like (different panel, but identical wiring)

marcmerlin:
@hzeller  Well, I will owe you a new matter soon, Vumapper :)

IMG_20200320_174205
And I have the back of this one
IMG_20200320_174436

hzeller:
You're really going to town with these panels :)

marcmerlin:
those old 64x32 panels that were donated to me, were gathering dust, I felt bad about it, and like many I have a few extra hours at home right now :)
Don't worry, after a not too long while, I'll get bored with this, and switch to another hobby like RC planes :)

rijotech:
I'm trying to replicate the panels setup but can't seen to find the right set of commands to get a straight image. @marcmerlin care to provide an example? I already haverything wired the same way with 12 64x32 panels.

marcmerlin:
@rijotech please try
~/rpi-rgb-led-matrix/examples-api-use/demo --led-gpio-mapping=regular --led-rows=32 --led-cols=64  --led-row-addr-type=0 --led-chain=5 --led-show-refresh --led-slowdown-gpio=1 --led-parallel=3   --led-panel-type=FM6126A --led-pwm-bits=7 --led-pwm-dither-bits=1 --led-pixel-mapper=V-Mapper:Z -D0
If that fails, please post picture of what you are seeing.

rijotech:
> @rijotech please try
> ~/rpi-rgb-led-matrix/examples-api-use/demo --led-gpio-mapping=regular --led-rows=32 --led-cols=64 --led-row-addr-type=0 --led-chain=4 --led-show-refresh --led-slowdown-gpio=1 --led-parallel=3 --led-panel-type=FM6126A --led-pwm-bits=7 --led-pwm-dither-bits=1 --led-pixel-mapper=V-Mapper:Z -D0
> If that fails, please post picture of what you are seeing.
> Obviously, your led chain will be 4 for 3x4 panels (mine was 5)

Thanks a lot @marcmerlin I already figured out I cloned your repo and figure it how to make it work. Thanks to you both! and stay with this pandemic.

lehni:
@marcmerlin is there a way to reverse the sequence of the chains, so I can go right to left, bottom to top in my wiring?

hzeller:
could you use use the mirror or rotate mappers stacked on the v-mapper ?

marcmerlin:
Henner's answer is best, my code was indeed only written for one direction, but hopefully the transforms on top will sort you out

lehni:
I've reversed the cable routing instead, all good :) Thank you both for the quick feedback!

PR #1013, opened 2020-03-21
Update gpio.h

Jeada91:
it seems some words was deleted in bool init (int...

### DIFF
--- a/include/gpio.h
+++ b/include/gpio.h
@@ -33,13 +33,13 @@ class GPIO {
 
   // Initialize before use. Returns 'true' if successful, 'false' otherwise
   // (e.g. due to a permission problem).
-  bool Init(int
+  bool Init(int slowdown) {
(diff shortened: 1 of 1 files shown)


hzeller:
The code is exactly as intended: there is a slowdown parameter, and it's default is either `RGB_SLOWDOWN_GPIO` (if defined) or 1 (if not defined).

Not sure what you intend to do, but your patch will essentially break the code.

Jeada91:
Sorry, yes you're rightIt's my mistake in copying lignes that lead to compilations error and Bad correction from meThanks for replying quicklyLe 21 mars 2020 07:43, Henner Zeller <notifications@github.com> a écrit :
The code is exactly as intended: there is a slowdown parameter, and it's default is either RGB_SLOWDOWN_GPIO (if defined) or 1 (if not defined).
Not sure what you intend to do, but your patch will essentially break the code.

PR #1014, opened 2020-03-21
Added V-Mapper:Z for flipped orientation panels chains

marcmerlin:
This allows using every other panel in a vertical chain, upside down
so as to be able to use the shortest HUB75 cables that come with
panels.
Instead of --led-pixel-mapper=V-Mapper use --led-pixel-mapper=V-Mapper:Z


marcmerlin:
IMG_20200321_152815
IMG_20200321_152804

marcmerlin:
My apologies for the git history, normally I'd blow my tree and start over fresh from yours to avoid all these merges with no file changes but CL entries, but I can't do that because I still have 2 outstanding pull requests that would get lost if I did that.

lib/pixel-mapper.cc
```diff
@@ -250,11 +258,29 @@ class VerticalMapper : public PixelMapper {
                                   int *matrix_x, int *matrix_y) const {
     int panel_width  = matrix_width  / chain_;
     int panel_height = matrix_height / parallel_;
+
     *matrix_x = (x % panel_width) +  int(y/panel_height)* panel_width;
     *matrix_y = (y % panel_height) + int(x/panel_width) * panel_height;
```
hzeller: Wouldn't it be easier and more readable to figure out if we need flipping first, and then do the one or the other ?

It is very hard to follow to first do the assignment, and then additional modifications later.

So start with an expression that figures out if we should do flipping
```
const bool needs_flipping = z_ && y / panel_height % 2 == 0;  // or something like that.
```

And then either some
```
if (needs_flipping) {
   *matrix_ = ...
   //
} else {
  *matrix_x = (x % panel_width) +  needs_flipping ? ... : int(y/panel_height)* panel_width;
  // ...
}
```

or, maybe even more readable
```
 *matrix_x =  (x % panel_width) + needs_flipping ? ... : ...
```

lib/pixel-mapper.cc
```diff
@@ -250,11 +258,29 @@ class VerticalMapper : public PixelMapper {
...
+
+    if (z_) {
+	int x_panel_offset_cnt = *matrix_x / panel_width;
+	int x_panel_offset = x_panel_offset_cnt * panel_width;
+	int y_panel_offset_cnt = *matrix_y / panel_height;
+	int y_panel_offset = y_panel_offset_cnt * panel_height;
+
+	if (x_panel_offset_cnt % 2) {
+
+	    *matrix_x = panel_width - (*matrix_x - x_panel_offset) - 1;
```
hzeller: here, you're overwriting calculation we already did above; we should make more clear that we either do the one thing or the other (see other comment above).

Also why not have everything in one expression instead of fixing up the value below, which makes it much harder to follow.
```
  *matrix_x = panel_width - (*matrix_x - x_panel_offset) - 1 + x_panel_offset;
```

lib/pixel-mapper.cc
```diff
@@ -250,11 +258,29 @@ class VerticalMapper : public PixelMapper {
...
+	    *matrix_y = panel_height - (*matrix_y - y_panel_offset) - 1;
+
+	    *matrix_x += x_panel_offset;
+	    *matrix_y += y_panel_offset;
+	}
+    }
   }
 
 private:
+  int z_;
```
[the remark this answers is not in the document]
marcmerlin: I'm actually used to boot/uint8_t/uint16_t/uint32_t coming from the arduino world, but I saw it was all ints here, and they all expand to 32bit registers, right?
Either way, yes, totally happy to be more specific

lib/pixel-mapper.cc
```diff
@@ -250,11 +258,29 @@ class VerticalMapper : public PixelMapper {
                                   int *matrix_x, int *matrix_y) const {
     int panel_width  = matrix_width  / chain_;
     int panel_height = matrix_height / parallel_;
+
     *matrix_x = (x % panel_width) +  int(y/panel_height)* panel_width;
     *matrix_y = (y % panel_height) + int(x/panel_width) * panel_height;
+
+    if (z_) {
+	int x_panel_offset_cnt = *matrix_x / panel_width;
```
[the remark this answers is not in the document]
marcmerlin: totally happy to, done. For my own learning: given that they can't be resolved and substituted at compile time, is const just being defensive, or does it actually help the code being more efficient?

lib/pixel-mapper.cc
```diff
@@ -250,11 +258,29 @@ class VerticalMapper : public PixelMapper {
...
+	int x_panel_offset = x_panel_offset_cnt * panel_width;
+	int y_panel_offset_cnt = *matrix_y / panel_height;
+	int y_panel_offset = y_panel_offset_cnt * panel_height;
+
+	if (x_panel_offset_cnt % 2) {
+
```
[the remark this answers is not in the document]
marcmerlin: oh yeah, left over from debugging stuff I removed, sorry (fixed)

lib/pixel-mapper.cc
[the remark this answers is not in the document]
hzeller: Readability mostly.

lib/pixel-mapper.cc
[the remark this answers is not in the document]
hzeller: yeah, int on the Pi are int32_t essentially. We don't care too much about size too much in this context. On microcontrollers I am also very specific about sizes.

lib/pixel-mapper.cc
marcmerlin: So, I'm a bit torn. I'll be honest that I had a hard time figuring out what U-mapper was doing, code that was doing smart things with virtually no comments :)
There are 2 issues maybe in conflict:
1) efficiency
2) readability / easy to follow the algorithm flow

For #1: I'll wager that the transform table is computed once at startup and that loosing maybe a millisecond (just made that up) to a few double assigns, is not a prime concern.

For #2: 
The function does 3 things
a) normal Vmapper
b) Z Vmapper on a panel that isn't flipped
c) Z Vmapper on a panel that is flipped

I wrote the codeflow to make it very clear which one of a b or c is happening.
By the time you make single line more complex assignments that do a b and/or c at the same time, it's pretty darn hard to figure out what's going on later.

I kind of like the code the way it is because it's pretty simple to follow.
The other part that would be confusing is that x_panel_offset_cnt is computed on the transposed value of X, after it's been mapped to the virtual array (i.e. the first assignment already happened). Sure, you can do smart math that would work before the assignment, but it seemed harder to write, and as a result even harder to re-read.

Keep in mind that we're not top SWEs, so keeping it a bit easier to read, is a plus IMO.
What do you think?

marcmerlin:
Thanks for the review. It feels like being at work :)

"The code looks like it evolved 'until wit worked' :)" => not really, I wrote it like this the first time, but maybe you mean that first the original transpose calculation happens, and then if there is a need to flip pixels, it does that afterwards.

lib/pixel-mapper.cc
marcmerlin: I'm pretty sure the optimizer makes it all the same in the end (i.e. it get collapsed) and it was slightly easier to read, but I agree that this one is not a big deal to collapse, so I just did so.
Obviously you can also do +2 * x_panel_offset instead of  -(- x_panel_offset) + x_panel_offset  but I wager that would be much less obvious to read, and the compiler should definitely do it for us.

lib/pixel-mapper.cc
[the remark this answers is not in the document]
marcmerlin: ok, I don't have that habit (probably I should), thanks for the info (as well as for bool)

lib/pixel-mapper.cc
hzeller: Yes, efficientcy we don't care, it only happens once.

 But readability. Multiple assignments to things require you to keep track of it when reading. You do and undo things, while all we need to choose is if we need flipping or not, and formulate the expression accordingly.

What happens now is
  * do one thing.
  * oh wait, we are actually doing the other thing .. maybe.
  * we actually need to do the other thing and overwrite which we did first

This is really hard to follow.
So this is why I suggest to change this to
   * do we need to do the other thing ?
   * Yep: do other thing
   * No: do one thing.

And once you've done this, you'll notice, that the ?: version of the same is even more readable.

lib/pixel-mapper.cc
```diff
@@ -237,24 +245,39 @@ class VerticalMapper : public PixelMapper {
...
                                   int *matrix_x, int *matrix_y) const {
-    int panel_width  = matrix_width  / chain_;
-    int panel_height = matrix_height / parallel_;
+    const int panel_width  = matrix_width  / chain_;
+    const int panel_height = matrix_height / parallel_;
+
     *matrix_x = (x % panel_width) +  int(y/panel_height)* panel_width;
     *matrix_y = (y % panel_height) + int(x/panel_width) * panel_height;
+
+    if (z_) {
```
hzeller: The cyclomatic complexity is still there because you have two nested ifs. But if you move the offset_cnt alculation out, you can have the if-condition in one expression

```c++
const int x_panel_offset_cnt = *matrix_x / panel_width;
if (z_ && x_panel_offset_cnt % 2 == 1) {
  ...
}
```

lib/pixel-mapper.cc
marcmerlin: Mmmh, so please don't take this as me trying to argue with you :)
The way I wrote it/meant it was:
a) transform from Hmapping to Vmapping
b) if V:Z, see if the resulting panel requires inverting
c) if so, deconstruct the X coordinate to figure out if it ends up on an odd or even panel. If it's an even panel, invert X and Y

a) needs to be done regardless, it's not a) or c), nor does c) undo a)
Combining a and c in one step can be done, but it definitely felt harder to write, and likely not understandable anymore when read.
That said, keep in mind that you're talking to a programmer who isn't as smart as you are :)

If you do see an obvious way to rewrite this in a short and yet still readable way, would you be ok merging this and modifying it the way you had in mind?
(I'll be happy to test your patch before you submit)

hzeller: (and yes, calculating from the already transposed value is particularly mind-bending)

hzeller: I disagree that a) needs to be done regardless. We either do a) or we do the 'flipped a'. Right now you have to take the one a) and have to do gymnastics to then flip it.

I don't try to make it more complicated or 'smart', I actually suggest this to make it more simple that even I would understand it (right now, it twists my brain).

We can merge it and I can have a look later.

examples-api-use/README.md
```diff
@@ -315,23 +317,31 @@ Here an example with 3 chains of 4 panels (128x64) for a total of about
...
 
- (This is also a good time to notice that such large arrangement is probably an
+ (This is also a good time to notice that 384x256 with 12 128x64 panels, is probably an
 upper limit of what you can reasonably output without having an unusable fresh
 rate (Try these options to help: --led-pwm-bits=7 --led-pwm-dither-bits=1 and get about 100Hz)).
 
-This can also be used to improve the refresh rate of a long display even
-if it is only one panel high (e.g. for a text running output) by
+This shows the wiring of a 3x5 Vmapper:Z array built by Marc MERLIN, using 15x 64x32 panels:
+![IMG_20200320_174436](https://user-images.githubusercontent.com/1369412/77215646-80838200-6ad2-11ea-977a-9d903933376c.jpg)
```
hzeller: if you check this into the toplevel `img/` directory, we can refer to it here with .`./img/`.
The githubusercontent link doesn't necessary look like it is stable.

lib/pixel-mapper.cc
marcmerlin: Sure, I can do that. It does create one extra division in the case of panels with no inversion, but we agreed that efficiency is not the main problem, so if you prefer it as a single if, done.
And I did try to move that const up with the other ones above. I know some people object to having variable declarations in the middle of a block, but if it is to remain a const, it's unavoidable.

examples-api-use/README.md
marcmerlin: They have been stable everywhere I've used them, they use a very long UUID, but I have no problem checking that in.
Done.

lib/pixel-mapper.cc
marcmerlin: Done

lib/pixel-mapper.cc
marcmerlin: "and yes, calculating from the already transposed value is particularly mind-bending" => you'll hate me, I found it easier to vizualize after the transposition than before :)
"We can merge it and I can have a look later." => 
that would be great. I am curious to see how it can be best written.

marcmerlin:
If you prefer this picture or would like both, I can replace or put both.
Plus: no power wires, so it's easier to show the Z and the panel orienation
Minus: doesn't show how power is distributed, given that you can't use the rails.
115_20200310_FastLED_RPIRGBPanel_GFX_192x160

hzeller:
Given that we want to illustrate the way the zig-zag wiring goes, this is a totally fine picture, no need to worry about power-connection ([img/chained-64x64](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/img/chained-64x64.jpg) doesn't have power connectors either).

Make sure to scale it down (~1024 wide ?) so that it is sufficiently small - because it will be part of the github repo that people check out on their Pi with possibly a slow connection.

If you want to see how the final rendered markdown looks like while editing it, I can recommend the [grip](https://github.com/joeyespo/grip) tool.

marcmerlin:
Sure, done.
Actually you can see the rendered version here 
https://github.com/marcmerlin/rpi-rgb-led-matrix/blob/Vmapper_Z/examples-api-use/README.md
but thanks for the grip link, didn't know about it. Normally I edit README.md files directly in the github interface which can render it online before you commit.

hzeller:
Cool, how can you render a markdown in github ?

hzeller:
(edit, I mean)

marcmerlin:
1) go to https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/README.md
2) click edit
3) click preview changes

hzeller:
Thanks, merged!

hzeller:
I've simplified the expressions to be just single assignment ( https://github.com/hzeller/rpi-rgb-led-matrix/commit/66fc98ddb156c9e6a0c5e879e26eed4fb45f394d ). This improves readability tremendously for me as it is now clear what happens and why.

Can you verify with your panel that it indeed behaves the same ?

PR #1023, opened 2020-03-23
secure radius-max

Jeada91:
Hello, just a little proposition to secure radius_max value that hang and reboot my raspberry PI because I have 64 cols*32 rows and radius_max was computed to 32 and this value is outside of my dot matrix display.

### DIFF
--- a/examples-api-use/minimal-example.cc
+++ b/examples-api-use/minimal-example.cc
@@ -30,7 +30,7 @@ static void DrawOnCanvas(Canvas *canvas) {
 
   int center_x = canvas->width() / 2;
   int center_y = canvas->height() / 2;
-  float radius_max = canvas->width() / 2;
+  float radius_max = fmin (canvas->width(),canvas->height()) / 2;
   float angle_step = 1.0 / 360;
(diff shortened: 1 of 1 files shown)


hzeller:
mmh, this should _never_ hang and crash the system as pixel coordinates should be checked if they are out of range.
If so, something else needs to be fixed. I'll have a look.

Jeada91:
effectively it seems my PI hangs for another reason because when circle expansion goes out the matrix panel it crash at different time and not exactly at first drawing outside my panel.
I'll do some other investigations
Thanks for your reply

Jeada91:
My problem isn't due to software failure but hardware failure. After several test I've orbserved that my RPI hangs with a lot of pixel iluminated independtly of pixels going out of the matrix.
As my RPI is on the same power supply (chineese 10A) than my DMD, power supply is not able to respond to the demand of current
By reducing birghtness all works well.

hzeller:
Cool finding. The power spikes are particularly harsh on power supplies; having separate wires from your matrix to the power supply and the pi to the power supply and capacitors directly at the matrix and Pi can help.

PR #1032, opened 2020-03-28
clean up c# bindings and add support for fonts

digital-synapse:


lib/led-matrix-c.cc
```diff
@@ -20,10 +20,14 @@
 #include <stdio.h>
```
hzeller, line 20: I don't understand the changes in this file, aren't they long there since f65aa06a031a3aaef7d0fee39db15015ee72c844 ?
 Is this an old branch that you need to rebase ?

hzeller:
Did this just pop up because you mentioned it in the NodeJS binding, or are there changes you need to add here.
It looks like it is based on a very old branch, and you already pushed the necessary changes in 2017 ?

In any case, I'd suggest a rebase so that it is mergeable and see what is needed.

digital-synapse:
Thanks. I will refresh and check. Mabey this PR isn't actually needed.

gingters:
@digital-synapse I'm currently working on my own (slightly biased) C# binding for .NET Core (not Mono): https://github.com/gingters/SharpPiLed

I did not yet publish a nuget package for that because it's still work in progress, but perhaps you still might to have a look at that and my updated examples.

I already ported all existing C# binding examples and am now working to port the native C++ examples (Scrolling text and all the other samples in the demo project).

While I do that I am also working on exposing more of the C++ api that is used in these examples for usage in my C# wrapper and making PRs here for that.

hzeller:
Should we close this now and have a fresh, rebased PR later ?
Maybe you can coordinate @digital-synapse and @gingters ?

marcmerlin:

PR #1040, opened 2020-04-02
support for P5 header on rpi1b+ model

lucky01:
added support for the P5 header on a rpi1b+ and designed a small pcb using this header

### DIFF
--- a/adapter/active-rpi1b+/rpi2dmd-pi1B-v2.brd
+++ b/adapter/active-rpi1b+/rpi2dmd-pi1B-v2.brd
@@ -1,0 +1,2498 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE eagle SYSTEM "eagle.dtd">
+<eagle version="7.4.0">
+<drawing>
+<settings>
+<setting alwaysvectorfont="no"/>
+<setting verticaltext="up"/>
+</settings>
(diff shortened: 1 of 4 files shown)


hzeller:
Mmh, this looks like a somewhat specialized form of connecting to the old Pis by using one of the additional headers, but even then this does not bring us over the 1 chain threshold, so it doesn't provide a huge advantage to the user (it might keep some pins free for other user, e.g. SPI).

I am reluctant to merge this, as introducing another variant of a single-chain interface for an Raspberry Pi not available on the market anymore might just confuse users.

(Also, the board should be in KiCAD format, so that this project can be used with OpenSource tools all the way down.)

lucky01:
I must admit that my pcb is designed for my specific usecase but it should be possible to add a second or even a third chain with the additional pins available by adding the P5 header. The problem was that I could not address these pins with your original source because you upper port address limit was set to 27 instead of 31.
Maybe you could just commit the changes removing this limit and maybe the pinmapping of my board that I can use the most current version of your code for my project. The eagle file can be edited with the trial version of eagle which has a limit of 100x100mm pcb size.

hzeller:
Alright, make P5 pins available in last commit.

marcmerlin:
apologies that this got dropped. If this still makes sense to add, please resubmit against master since it won't merge anymore

PR #1041, opened 2020-04-05
Add font height and baseline to exported methods

gingters:
Add additional methods of the font class to the exported functions in in the .so so that they can be consumed in the C# wrapper.


hzeller:
Thanks!

gingters:
Lol! I never thought that would be so fast. Awesome! Thanks! That makes my binding work a LOT easier! ;)

PR #1043, opened 2020-04-06
Add outline font creation to exported methods

gingters:
Also export the create_outline_font method to the .so for consumption by the wrappers.


hzeller:
Thanks (sorry for the delay, I was busy with other things in that time and didn't see the pull request)

PR #1054, opened 2020-04-15
Added argument for scroll rate

ElBertrando:
Similar to c++ examples for scrolling-text-example -- #TODO -- Add scroll rate to API so that this argument can be added

### DIFF
--- a/bindings/python/samples/runtext.py
+++ b/bindings/python/samples/runtext.py
@@ -9,7 +9,7 @@ class RunText(SampleBase):
     def __init__(self, *args, **kwargs):
         super(RunText, self).__init__(*args, **kwargs)
         self.parser.add_argument("-t", "--text", help="The text to scroll on the RGB LED panel", default="Hello world!")
-        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=int)
+        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=str)
         #self.parser.add_argument("--led-panel-type", action="store", help="Needed to initialize special panels. Supported: 'FM6126A'", default="", type=str)
 
     def run(self):
@@ -19,11 +19,11 @@ class RunText(SampleBase):
         textColor = graphics.Color(255, 255, 0)
         pos = offscreen_canvas.width
         my_text = self.args.text
-        scrollSpeed = self.args.speed
+        initSpeed = self.args.speed
 
         while True:
             offscreen_canvas.Clear()
-            len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, my_text, scrollSpeed)
+            len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, my_text, initSpeed)
             pos -= 1
             if (pos + len < 0):
                 pos = offscreen_canvas.width
--- a/bindings/python/samples/runtext.py
+++ b/bindings/python/samples/runtext.py
@@ -9,7 +9,7 @@ class RunText(SampleBase):
     def __init__(self, *args, **kwargs):
         super(RunText, self).__init__(*args, **kwargs)
         self.parser.add_argument("-t", "--text", help="The text to scroll on the RGB LED panel", default="Hello world!")
-        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=str)
+        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=int)
         #self.parser.add_argument("--led-panel-type", action="store", help="Needed to initialize special panels. Supported: 'FM6126A'", default="", type=str)
 
     def run(self):
@@ -19,11 +19,11 @@ class RunText(SampleBase):
         textColor = graphics.Color(255, 255, 0)
         pos = offscreen_canvas.width
         my_text = self.args.text
-        initSpeed = self.args.speed
+        scrollSpeed = self.args.speed
 
         while True:
             offscreen_canvas.Clear()
-            len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, my_text, initSpeed)
+            len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, my_text, scrollSpeed)
             pos -= 1
             if (pos + len < 0):
                 pos = offscreen_canvas.width
--- a/bindings/python/samples/runtext.py
+++ b/bindings/python/samples/runtext.py
@@ -9,7 +9,7 @@ class RunText(SampleBase):
     def __init__(self, *args, **kwargs):
         super(RunText, self).__init__(*args, **kwargs)
         self.parser.add_argument("-t", "--text", help="The text to scroll on the RGB LED panel", default="Hello world!")
-        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=str)
+        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=int)
         #self.parser.add_argument("--led-panel-type", action="store", help="Needed to initialize special panels. Supported: 'FM6126A'", default="", type=str)
 
     def run(self):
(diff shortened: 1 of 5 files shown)


hzeller:
You have added a flag, but have not implemented the functionality. You probably want to update the sleep time depending on your parameter.

ElBertrando:
I added the same logic as the line of code above for "--text". Do I need to add more logic for the functionality ? I have search the samplebase.py code and the core.pyx code and see not reference for the logic behind the "--text"..

ElBertrando:
Hi Henner,
In which file to do I need to implement the functionality and update the sleep time ? In "runtext.py" or "samplebase.py" ? Again, it appears I add the argument in the runtext class, which I did as you did for "text" but I do not see any more code for text if I am using that as the model to add sleep time.

ElBertrando:
Modified DrawText in text-scroller.cc, led-matrix.cc, graphics.cc, and graphics.h to add argument for speed when using graphics.DrawText() in runtext.py  -- Please review to check if I performed these changes properly

hzeller:
You don't need speed in the parameter of function that prints the text. This function never moves the text. There is nothing that would change anything here.

The scrolling happens by starting at a [different position in each animation phase](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/bindings/python/samples/runtext.py#L24). So all you need to do to modify the speed is have a [different sleep time](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/bindings/python/samples/runtext.py#L28)  between animation phases.

ElBertrando:
I modified for an an input variable -- Hopefully this is the right track for the run function.

bindings/python/samples/runtext.py
```diff
@@ -9,23 +9,27 @@ class RunText(SampleBase):
...
+        if(s > 0.0):
             offscreen_canvas.Clear()
             len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, my_text)
             pos -= 1
             if (pos + len < 0):
                 pos = offscreen_canvas.width
 
-            time.sleep(0.05)
+            #time.sleep(0.05)
+            time.sleep(s)
```
hzeller, line 32: Just use `time.sleep(1.0 / speed)`

bindings/python/samples/runtext.py
```diff
@@ -9,23 +9,27 @@ class RunText(SampleBase):
     def __init__(self, *args, **kwargs):
         super(RunText, self).__init__(*args, **kwargs)
         self.parser.add_argument("-t", "--text", help="The text to scroll on the RGB LED panel", default="Hello world!")
+        self.parser.add_argument("-s", "--speed", help="Approximate letters per second.", default="-s0",type=float)
+        #self.parser.add_argument("--led-panel-type", action="store", help="Needed to initialize special panels. Supported: 'FM6126A'", default="", type=str)
 
-    def run(self):
+    def run(self, s):
```
hzeller, line 15: If you now pass the speed as paremter to run, you have to call run with this parameter, but this is not expected.

Instead, just read the speed parameter from the flag (as you did before) and calculate the time by doing something like `1.0 / speed`. So if speed is 20, then you get 0.05 (as before), but if speed is 10, you get (0.1 ... longer time wait).

bindings/python/samples/runtext.py
```diff
@@ -9,23 +9,27 @@ class RunText(SampleBase):
...
         offscreen_canvas = self.matrix.CreateFrameCanvas()
         font = graphics.Font()
         font.LoadFont("../../../fonts/7x13.bdf")
         textColor = graphics.Color(255, 255, 0)
         pos = offscreen_canvas.width
         my_text = self.args.text
 
-        while True:
+        #while True:
+        if(s > 0.0):
```
hzeller, line 24: What is the purpose of this change ?

bindings/python/samples/runtext.py
ElBertrando: Some times I  want the speed to be 0, and sometimes have it run across the panel. If I use time.sleep(1.0/speed),then I can't divide by 0. In my situation, I plan to call this function with different text and different speeds (although the speed will be 0 most of the time).

bindings/python/samples/runtext.py
ElBertrando: I plan on calling the run function based on different scenarios, so I want to control the speed of the text - ranging from 0 and greater than 0.

bindings/python/samples/runtext.py
hzeller: The `run()` function is called from the thread in samplebase.py, so you need to keep it wit no parameters.

If speed is zero, you essentially simply don't want any movement, so in that case, don't update `pos`. In the sleep function you can then do something like `if speed == 0 then 10 else 1.0 / speed`.

(Also important to consider: if you don't have any scrolling, you probably want to move the initial pos to a place where it is visible)

bindings/python/samples/runtext.py
hzeller: the run function is the function running in an eternal loop in the thread. If you want to change that, you probably want to rename the funcion.

ElBertrando: How about I just write a new function that take a boolean value to perform the while loop like this 
```
    def runTextWithSpeed(self, speedOfText, textToDisplay, textColorToDisplay, displaySomething):
        while(displaySomething):
            offscreen_canvas = self.matrix.CreateFrameCanvas()
            font = graphics.Font()
            font.LoadFont("../../../fonts/7x13.bdf")
            #textColor = graphics.Color(255, 255, 0)
            textColor = textColorToDisplay
            pos = offscreen_canvas.width
            thisText = textToDisplay

            if(speedOfText == 0):
                pos = 20
                offscreen_canvas.Clear()
                graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, thisText)
            elif(speedOfText > 0):
                offscreen_canvas.Clear()
                len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, thisText)
                pos -= 1
                if (pos + len < 0):
                    pos = offscreen_canvas.width
                time.sleep(1.0/speedOfText)
                offscreen_canvas = self.matrix.SwapOnVSync(offscreen_canvas)
        else:
            offscreen_canvas.Clear()
```

hzeller: here, you're creating a frame canvas and load the font in every loop iteration. You'll run out of memory soon.

ElBertrando: Ok I see that -- -I've modified the function to this:
```
    def runTextWithSpeed(self, speedOfText, textToDisplay, textColorToDisplay, displaySomething):
        if(displaySomething):
            offscreen_canvas = self.matrix.CreateFrameCanvas()
            font = graphics.Font()
            font.LoadFont("../../../fonts/7x13.bdf")
            #textColor = graphics.Color(255, 255, 0)
            textColor = textColorToDisplay
            pos = offscreen_canvas.width
            thisText = textToDisplay

            if(speedOfText == 0):
                pos = 20
                offscreen_canvas.Clear()
                graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, thisText)
            elif(speedOfText > 0):
                while True:
                    offscreen_canvas.Clear()
                    len = graphics.DrawText(offscreen_canvas, font, pos, 10, textColor, thisText)
                    pos -= 1
                    if (pos + len < 0):
                        pos = offscreen_canvas.width
                    time.sleep(1.0/speedOfText)
                    offscreen_canvas = self.matrix.SwapOnVSync(offscreen_canvas)
        else:
            offscreen_canvas.Clear()
```

ElBertrando: Hi Henner,
Also when I create a an offscreen_canvas, Do I need to pass in the number of rows and columns as parameters ? I traced the it back to the "led-matrix.cc" where frame-buffer arguments are passed in. 
Thanks

hzeller: The matrix creates the frame canvas internally, so it will return you an instance with everything properly set.

ElBertrando: Thx  -- Well if  my mofifed code looks OK to you then I suppose there is no need to proceed further with this pull request because I can simply add this function to my project.

hzeller: Perfect!
Have fun with your project and LEDs!

PR #1055, opened 2020-04-16
Added support for Compute Module 6 parallel chains

arahasya:
Uses up all available GPIO from GPIO2 - GPIO45

GPIO operations to uint64 for GPIO > 31


arahasya:
As I was researching about the GPIO pins got this from the datasheet

**Each GPIO can supply up to 16mA, aggregate current per bank must not exceed 50mA**

What is the o/p current of each pin right now? For 6 chains we are at least using 42-44 pins
I hope we are below the limit :)

arahasya:
Hi Henner,
I have now tested Custom compute module PCB and am awaiting delivery of active-6 adapter. Maybe you would like to decide to include this in the library? I can also provide links to sell the hardware.
Let me know. Thanks!

hzeller:
I would like to include compute module compatibility, but I need to have a test-setup.
Do you have a link to your open source printed circuit board ?

arahasya:
Do you mean the Compute module PCB? It is not open source yet. We are still deciding on that because we haven't developed the PCB for this library specifically. It is a general Raspberry pi Compute module development board which we are going to sell for various other applications too. Also, it is unfeasible to manufacture in small quantities as it is a 4-layer board. The BOM count is 44. So unless the production is in 100's the cost is slightly more the cost of 2 pi's. 
I can mail you the hardware for your test setup in a few days when we have produced more. In the latest revision I have added CM108B usb sound driver chip on the PCB itself. Have to test it out now

hzeller:
I see. So your active-6 one is also specific to your compute module development board ?
If so, we should probably build a active-6 board for the standard compute module coming from the Raspberry Pi foundation as a starting point. I'll have a look at that.

I have merged your pull request, but until I am able to verify it with hardware, it is in the compute-module branch:
https://github.com/hzeller/rpi-rgb-led-matrix/tree/compute-module
Once things work on there, I'll merge it to master.

hzeller:
I am a bit concerned about more memory usage, as the framebuffer now would store 64 bit values, which makes it put more pressure on caches and possibly processing time (and disk space if used with the 'stream' option).
I am thinking of some solutions.

hzeller:
I've now made it a [compile time option in the Makefile](https://github.com/hzeller/rpi-rgb-led-matrix/blob/compute-module/lib/Makefile#L132)
```
DEFINES+=-DENABLE_WIDE_GPIO_COMPUTE_MODULE
```

.. in the course of that, thinned out the public API to not leak the GPIO internals anymore to users.

arahasya:
> I see. So your active-6 one is also specific to your compute module development board ?
Yes.
> If so, we should probably build a active-6 board for the standard compute module coming from the Raspberry Pi foundation as a starting point. I'll have a look at that.
> I am not sure of that approach. The active-6 board for the Official development board will require a connections of at least 100 pins because of the GPIO distribution. Also the cost is 99$. 

> Once things work on there, I'll merge it to master.
Thanks!

arahasya:
Here is a working prototype -
WhatsApp Image 2020-08-04 at 10 55 54

Here is the revised newer version with on-board sound card -

[3D.PDF](https://github.com/hzeller/rpi-rgb-led-matrix/files/5020235/3D.PDF)

I am in process to produce the next version. But I can mail you the first version with the active-6 board to save you some time.

arahasya:
Also the advantage with the above board is it is pin compatible with regular pi
Which makes it compatible for all pi HATs and active 3 board. 
Compute module Dev board is not compatible

hzeller:
Your board looks quite nice! Do you just the first part of the GPIO the same as the Raspberry Pi, then extend it from there ?
You find my email address on my github profile; if you send me an email, I'll give you my post address.

arahasya:
Yes, the GPIO is same and just extended but it starts from the right side and not the left. (for routing purposes)

So it is still compatible with RPi 40-pins but the HAT will be outward instead on inward. This was also a decision taken considering the event of change of compute module you should not have to remove the HAT. Removing and plugging in HATs in the pins becomes very difficult if the pins get bend.

Another advantage worth mentioning is the board we designed has Ethernet in built and 2 USB ports. Compute Module dev board has only one USB port and no Ethernet/wifi. So you must use Usb to ethernet to ssh into the pi.

hzeller:
FYI, I've merged this into the main branch now (but couldn't test it yet of course).
Requires to set the `ENABLE_WIDE_GPIO_COMPUTE_MODULE` define in [`lib/Makefile`](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/Makefile#L138)

arahasya:
> I've now made it a [compile time option in the Makefile](https://github.com/hzeller/rpi-rgb-led-matrix/blob/compute-module/lib/Makefile#L132)
> 
> ```
> DEFINES+=-DENABLE_WIDE_GPIO_COMPUTE_MODULE
> ```
> 
> .. in the course of that, thinned out the public API to not leak the GPIO internals anymore to users.
> 
> So except for reading values, this removes the ability for the user to set values via GPIO. I might have to revisit this.


How to change the GPIO mapping now? Where is the actual assignment?

hzeller:
There are functions on the LEDMatrix which now allow to access GPIO for writing:

https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/include/led-matrix.h#L300-L309

Issue #1045, opened 2020-04-10
Discussion: expose argument parsing?


gingters:
Hi @hzeller,

I'd like to discuss exposing the logic that does the argument parsing through the C api for my C# binding. Reason being I am currently in the process of porting all the examples to C#, and at the moment I don't have all information available.

For example, if I start my sample with `--led-chain=2`, then the library does parse this, but the changes are not reflected back into the options (which get copied over to the struct). The scrolling text example needs the chain length information and the C++ version gets that from the updated matrix_options after parsing.

I did some experiments (as I already mentioned somewhere else I am really no C/C++ dev, but I seem to find the correct spots), and I like to propose the following changes:

1. I would add the RuntimeOptions struct to the C api  
  then we can set the gpio slowdown, daemon, drop privileges and init gpio also from C# code without the need to pass them along as injected command line args
2. I would like to expose the ParseOptionsFromFlags() method, via wrapper that takes the arguments and the two structs (options & runtime options)

For that to work, in my current draft, I re-used some of the code that copies the C to the C++ struct and back and copied the method for the new runtime options struct.

Then in the binding code I can create an options object which maps the properties to the two interop structs, and call the parse method from there.
The library will parse the command line, update the structs, and so my C# options object can read the updated values back. So my code knows what settings are set in the library, without the need to manually duplicate the parsing code in C#.

Would you, in general, be willing to accept such a PR?
With further discussion about implementation details, of course. When I put the struct-copy code into a separate function I got a warning about a unused argument, and couldn't figure out why, so I disabled that warning 🙈. Maybe you would be willing to give some hints on how to improve that, too?

hzeller:
(sorry, I was busy the last couple of weeks with some technical Covid19 responses, will try to get back to you asap)

gingters:
Absolutely no problem. No hurry here. ;)
Real live always goes first and hobbies are a bit of lower priority.

TDHolmes:
I'm also interested in having the `RuntimeOptions` struct added to the C API for my rust bindings

PR #1141, opened 2020-09-07
Expose runtime options to C API

TDHolmes:
In other language bindings (rust in my case), I'd like to fully control how the matrix is initialized without relying on argc/argv, as rust doesn't even give an easy way to access this C-like representation of command line input. Rather, I'd like to do my own command line parsing and fully control the options.

I think this might fulfill the needs of #1045 (cc. @gingters)

Breakdown of changes:
- add RGBLedRuntimeOptions struct to C header
- add `static_assert`'s to make sure C/C++ structures are at least the same size
- Add new C API that creates an `RGBLedMatrix` just from the given options & runtime options if they're non-NULL, without argc/argv at all
- Adds missing `do_gpio_init` initializer to C++ `RuntimeOptions` constructor
- fix apparent bug/typo in the copy back of options (`matrix_options` vs `default_opts` in `led_matrix_create_from_options_optional_edit`)

### DIFF
--- a/include/led-matrix-c.h
+++ b/include/led-matrix-c.h
@@ -154,6 +154,43 @@ struct RGBLedMatrixOptions {
 };
 
 /**
+ * Runtime options to simplify doing common things for many programs such as
+ * dropping privileges and becoming a daemon.
+ */
+struct RGBLedRuntimeOptions {
+  int gpio_slowdown;    // 0 = no slowdown.          Flag: --led-slowdown-gpio
+
+  // ----------
+  // If the following options are set to disabled with -1, they are not
+  // even offered via the command line flags.
+  // ----------
+
+  // Thre are three possible values here
+  //   -1 : don't leave choise of becoming daemon to the command line parsing.
+  //        If set to -1, the --led-daemon option is not offered.
+  //    0 : do not becoma a daemon, run in forgreound (default value)
+  //    1 : become a daemon, run in background.
+  //
+  // If daemon is disabled (= -1), the user has to call
+  // RGBMatrix::StartRefresh() manually once the matrix is created, to leave
+  // the decision to become a daemon
+  // after the call (which requires that no threads have been started yet).
+  // In the other cases (off or on), the choice is already made, so the thread
+  // is conveniently already started for you.
+  int daemon;           // -1 disabled. 0=off, 1=on. Flag: --led-daemon
+
+  // Drop privileges from 'root' to 'daemon' once the hardware is initialized.
+  // This is usually a good idea unless you need to stay on elevated privs.
+  int drop_privileges;  // -1 disabled. 0=off, 1=on. flag: --led-drop-privs
+
+  // By default, the gpio is initialized for you, but if you run on a platform
+  // not the Raspberry Pi, this will fail. If you don't need to access GPIO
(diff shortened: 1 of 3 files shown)


TDHolmes:
This worked on my setup via my rust language bindings. Interestingly, it compiled successfully on the rpi without the `#include <stdbool.h>` that the CI just failed with. I guess it must have been included somewhere else in the rpi's stdlib 🤷‍♂️

lib/led-matrix-c.cc
```diff
@@ -22,6 +22,9 @@
 #include "led-matrix.h"
 #include "graphics.h"
 
+// Make sure C++ is in sync with C
+static_assert(sizeof(rgb_matrix::RGBMatrix::Options) == sizeof(RGBLedMatrixOptions));
```
hzeller, line 26: Nice, this is a good idea!

lib/options-initialize.cc
```diff
@@ -38,7 +38,8 @@ RuntimeOptions::RuntimeOptions() :
   gpio_slowdown(1),
 #endif
   daemon(0),            // Don't become a daemon by default.
-  drop_privileges(1)    // Encourage good practice: drop privileges by default.
+  drop_privileges(1),   // Encourage good practice: drop privileges by default.
+  do_gpio_init(true)
```
hzeller, line 42: Oh, thanks, that was missing indeed.

lib/led-matrix-c.cc
```diff
@@ -98,14 +106,10 @@ static struct RGBLedMatrix *led_matrix_create_from_options_optional_edit(
...
     }
-  } else {
-    // CL args not provided. Assume defaults for RT options
-    runtime_opt.drop_privileges = 1;
-    runtime_opt.do_gpio_init = true;
   }
 
   if (opts) {
-#define ACTUAL_VALUE_BACK_TO_OPT(o) opts->o = default_opts.o
+#define ACTUAL_VALUE_BACK_TO_OPT(o) opts->o = matrix_options.o
```
hzeller, line 112: Do'h - I clearly never tested that :/ thanks for the fix!

include/led-matrix-c.h
```diff
@@ -153,6 +154,43 @@ struct RGBLedMatrixOptions {
...
+
+  // Drop privileges from 'root' to 'daemon' once the hardware is initialized.
+  // This is usually a good idea unless you need to stay on elevated privs.
+  int drop_privileges;  // -1 disabled. 0=off, 1=on. flag: --led-drop-privs
+
+  // By default, the gpio is initialized for you, but if you run on a platform
+  // not the Raspberry Pi, this will fail. If you don't need to access GPIO
+  // e.g. you want to just create a stream output (see content-streamer.h),
+  // set this to false.
+  bool do_gpio_init;
```
hzeller, line 191: In other places in the C structs, I just used `char` (e.g. `show_refresh_rate`) to behave as a bool, but I like using stdbool better.
If you want, I am happy for a separate pull request fixing the `char` to `bool` in the `struct RGBLedMatrixOptions`

hzeller (approved):

hzeller:
Thanks!

TDHolmes:
Thanks for being responsive and for making a sweet library!

greatballoflazers:
See #1362

PR #1148, opened 2020-09-12
2x Faster Python rotating block generator sample.

westonal:
The existing sample runs on my system at < 10 fps. I knew the python `SetPixel` fill rate was bad, but on further investigation a great deal of this is down to `sin` `cos` and the color mapping.

These improvements let it run at 20 fps. (The timing code is not included but just to be clear I'm not talking about the `--led-show-refresh` output).

- [x] Calculate sin and cos per frame, not per-pixel
  - [x] Then pre-calculate all sin and cos values for `range(0, 360)`
- [x] Pre-calculate all colors for `range(int(min_rotate), int(max_rotate))`

This required some in-lining, so I added comments

### DIFF
--- a/bindings/python/samples/rotating-block-generator.py
+++ b/bindings/python/samples/rotating-block-generator.py
@@ -11,6 +11,19 @@ def scale_col(val, lo, hi):
     return 255 * (val - lo) / (hi - lo)
 
 
+def rotate_about_point(x, y, cent_x, cent_y, sin, cos):
+    # translate so that the origin is now (cent_x, cent_y)
+    x -= cent_x
+    y -= cent_y
+    # rotate about the new origin
+    new_x = x * cos - y * sin
+    new_y = x * sin + y * cos
+    # translate back
+    new_x += cent_x
+    new_y += cent_y
+    return new_x, new_y
+
+
 class RotatingBlockGenerator(SampleBase):
     def __init__(self, *args, **kwargs):
         super(RotatingBlockGenerator, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


bindings/python/samples/rotating-block-generator.py
```diff
@@ -34,25 +29,52 @@ def run(self):
...
 
+            sin = sin_table[rotation]
+            cos = cos_table[rotation]
+
             for x in range(int(min_rotate), int(max_rotate)):
                 for y in range(int(min_rotate), int(max_rotate)):
-                    ret = self.rotate(x - cent_x, y - cent_x, deg_to_rad * rotation)
-                    rot_x = ret["new_x"]
-                    rot_y = ret["new_y"]
+                    # translate so that the origin is now (cent_x, cent_y)
```
hzeller, line 58: This assignment to temp variables looks bit awkward here and hinders readability; I'd probably make a function to more emphasize what is happening here.
It would get `x`, `y` and pre-calculated `sin`, `cos` values returns a tuple with the new coordinates; roughly something like this:

```Python
new_x, new_y = rotate_coordinates(x - cent_x, y - cent_y, sin, cos)
new_x += cent_x
new_y += cent_y
```

The implementation of that function is of course trivial. But since this is an example, we should also make sure to suggest good coding practices for people starting out with programming.

hzeller (changes requested):
Nice!
I mostly only have one suggestion to simplify readability.

bindings/python/samples/rotating-block-generator.py
westonal: OK, I've moved to a method. It cost a few fps, about 17-18 now on my setup. I made the method `rotate_about_point` as that's more useful and gave me a place to explain the subtraction and addition operations that otherwise might appear magical.

I decided not to build the sin/cos lookup table. one sin and cos operation per frame makes no difference and simplifies the example. It still teaches the principle of pre-calculating through the colours, while not breaking if someone wants to do sub degree increments.

hzeller:
Very cool, thank you. Merged.

hzeller:
I did have to do a little change that it rotates as before (due to some confusing way the original code was setting up the center)

PR #1151, opened 2020-09-14
PR for GPIO Pin-out change for compute-module mapping

arahasya:
Hi,

I have now successfully tested the active-6 adapter also. Changing the pin-out according to the adapter.
Had to use GPIO0 and GPIO1 because GPIO31 is used for LAN_RUN.
 Add 'force_eeprom_read=0' in '/boot/config.txt' to use GPIO0 GPIO1 without any issues.

[active-6.zip](https://github.com/hzeller/rpi-rgb-led-matrix/files/5217541/active-6.zip)

Here's a test video. I have 4 parallel chain display. In the video I have connected first, second, fifth and sixth chain to demonstrate the output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ The LED-matrix library is (c) Henner Zeller <h.zeller@acm.org>, licensed with
 source and all your modifications available to the receiver of such product so
 that they have the freedom to adapt and improve).
 
+__If you'd like help, please do not file a bug, use the discussion board instead: 
+https://rpi-rgb-led-matrix.discourse.group/__.
+
 Overview
 --------
 The RGB LED matrix panels can be scored at [Sparkfun][sparkfun],


KentWalker:
HI,

When you say GPIO31 is used for LAN_RUN, is it a dedicated function of the GPIO pin that can't be changed?
I'm designing a board that uses the CM3+ module directly, so no ethernet.

arahasya:
> HI,
> 
> When you say GPIO31 is used for LAN_RUN, is it a dedicated function of the GPIO pin that can't be changed?
> I'm designing a board that uses the CM3+ module directly, so no ethernet.

You can use GPIO31 for other purpose also. It isn't specific for LAN_RUN. But it was the default LAN_RUN in some versions so I used it. 

How will you connect to the pi? Wireless adapter? HDMI?

KentWalker:
Hi @arahasya,
Communications to the CM3+ is via RS232, the displays I use are large message boards so not a lot of data traffic is required and the message controller is not more than 10 metres from the display.  I set the CM3+ up initially in a dev kit and once the serial port is configured, any other changes can be done via a terminal program.

arahasya:
Can you tell me the resolution and scanning of your matrix? You might not need a CM3+ if you just need to display messages

KentWalker:
I'm already using CM3 modules as interfaces as no SD card required.  Our biggest display (so far) is 192w x 96h pixels (6 panels wide by 6 panels high) P10 1/2 scan outdoor and using 3 parallels is pushing it.  Resolution isn't so much the issue, the displays are visible from over 30 metres (100 feet) but problems start with longer chains.  Besides, it's better to have a generic controller board that can provide an interface to 1-6 parallel chains, it costs not much more than one that does 1-3 parallel chains.

arahasya:
So you are using 33x32 panels of 1:2 right?
You can easily use U-mapper to convert 3 chains to 6 chains. Reduce pwm-bits as you are displaying messages only.

You really don't require CM+. As for SD card you can boot from pen drive through USB port

arahasya:
Tried your setting to check refresh rate.

956.9Hz (lowest: 618.0Hz)

It is very high. You definitely don't need to use CM3 +

hzeller:
Could you make the pull request relative to current head ?

I suspect since the rest is already enabled for CM (when `ENABLE_WIDE_GPIO_COMPUTE_MODULE` defined in [lib/Makefile](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/Makefile#L138)), you only need to change [`hardware-mapping.c`](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/hardware-mapping.c#L221) to enable the final pin mapping.

Having a definitive pin mapping would make sense once we have an active 6 connector available that people can use, not just a private project. So we should have first a Kicad project in the [adapter/](https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/adapter) directory and with it fix the pin-mapping to that. Do you have a Kicad project ready to contribute @arahasya or @KentWalker ?

arahasya:
[PEBPL117.zip](https://github.com/hzeller/rpi-rgb-led-matrix/files/5323601/PEBPL117.zip)

Issue #1170, opened 2020-10-05
Creating RGBMatrix with Python bindings while running as root hijacks/disables file system IO


BenjaminPelletier:
I have a basic Python program modeled off of the `runtext.py` binding example, and I've copied one of the fonts (`7x13.bdf`) into a `fonts` subfolder relative to my script (`display_driver.py`).  It works when the font is owned by `pi`:

```
pi@poolpi:~/poolmonitor/display_driver $ sudo chown -R pi:pi fonts
pi@poolpi:~/poolmonitor/display_driver $ python3 display_driver.py
Press CTRL-C to stop
FYI: not running as root which means we can't properly control timing unless this is a real-time kernel. Expect color degradation. Consider running as root with sudo.
Can't set realtime thread priority=99: Operation not permitted.
        You are probably not running as root ?
        This will seriously mess with color stability and flicker
        of the matrix. Please run as `root` (e.g. by invoking this
        program with `sudo`), or setting the capability on this
        binary by calling
        sudo setcap 'cap_sys_nice=eip' $THIS_BINARY
^CExiting
```
(the RGB matrix is displaying text properly until I hit ctrl-c)

However, my problem is that LoadFont has an error when I run as root:
```
pi@poolpi:~/poolmonitor/display_driver $ sudo python3 display_driver.py
Press CTRL-C to stop
Traceback (most recent call last):
  File "display_driver.py", line 51, in <module>
  File "display_driver.py", line 34, in loop
SystemError: <method 'LoadFont' of 'rgbmatrix.graphics.Font' objects> returned NULL without setting an error
```

I thought this was likely due to the font file being owned by `pi`, so I tried changing ownership to `root`:
```
pi@poolpi:~/poolmonitor/display_driver $ sudo chown -R root:root fonts
pi@poolpi:~/poolmonitor/display_driver $ python3 display_driver.py
Press CTRL-C to stop
FYI: not running as root which means we can't properly control timing unless this is a real-time kernel. Expect color degradation. Consider running as root with sudo.
Can't set realtime thread priority=99: Operation not permitted.
        You are probably not running as root ?
        This will seriously mess with color stability and flicker
        of the matrix. Please run as `root` (e.g. by invoking this
        program with `sudo`), or setting the capability on this
        binary by calling
        sudo setcap 'cap_sys_nice=eip' $THIS_BINARY
Traceback (most recent call last):
  File "display_driver.py", line 51, in <module>
    loop()
[...]

PR #1176, opened 2020-10-12
Add API documentation to Python bindings

BenjaminPelletier:
This is primarily in response to #1170.  Also, this PR fixes a few minor grammar issues in the documentation.

### DIFF
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -80,12 +80,12 @@ swap with `SwapOnVSync()` (this is the fastest method).
 Using the library
 -----------------
 
-Be aware of the fact, that using the RGBMatrix requires root privileges.
-Therefore you will need to run all you python scripts as using sudo.
+Be aware of the fact that using the full performance of the RGBMatrix requires root privileges.
+Therefore you should run all you python scripts as using `sudo`.
 
-You find examples in the [samples/](./samples) subdirectory.
+You may find examples in the [samples/](./samples) subdirectory.
 The examples all use the [samplebase.py](./samples/samplebase.py) that provides
-some utility to all example programs, such as command-line parsing: all
+some basic capabilities to all example programs, such as command-line parsing: all
 sample-programs accept `--led-rows`, `--led-chain` and `--led-parallel` as
 command line options to adapt to your configuration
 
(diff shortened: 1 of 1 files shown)


hzeller:
Thanks.

daveythacher:
Just to be clear this security measure is only as good as the application that calls this library. If there is a noticeable delay in the call to init the library/matrix, bad stuff can still happen. At that point full privilege's will exist. This code base would likely fail or be blocked on critical systems following a security review.

I suspect the reason for why it does this has to do with performance. One could isolate the library into its own process, but communication between processes may prove degrading. A front end library would also likely exist for application code. This would prevent application code from running in super user state, unless required. Basically creating a user space driver or following a microkernel approach.

Now one could manually extent this functionality using the existing functionality.

hzeller:
This is why people are given the choice of doing the privilege dropping themselves. But the default tries to be as useful as possible for the user who doesn't know how to do that.

daveythacher:
I think this should be backed out. It is not technically correct. Even if running as daemon and file is owned by Pi, the file could still be accessed if global access permission are set correctly. This file is not likely critical enough to be blocked from other users. If it is critical the code could be changed to make it jump to Pi instead of daemon.

However recommending running an application as super user all the time is probably critical.

daveythacher:
Technically Linux would expect you do hardware operations in the kernel then communicate via a sysfs like interface. That could be slow, however it provides division of privilege's in the manner at the core of Linux security. This could isolate the need for root out of user space and more importantly the library.

Overall if you use a better notion of hardware you can avoid these problems. For example you can grant access to things like SPIDEV, which will allow any process running under those users to have full control over devices connected to bus. However the kernel could still restrict this some.

For this project there is still issues. This process just maps into /dev/mem for IO and PWM control. A kernel driver could replace this. You can make an interface which work at filesystem however complexity can rise there. Overall the kernel is supposed to manage hardware. However these notions create overhead which are problematic for this approach.

This project shows a poor understand standing of user space vs kernel space. It also shows a poor understanding of embedded devices. Proceeding with this notion will create issues. Software centric design is not stable. Technically what you are trying to do is possible. In certain cases this project will work. It is the library part that actually really invites issues. 

However not exactly. Overall what causes issues is the context of existence. It is undefined, and therefore can be defended and attacked at will. There is a way to make stuff like this work, however the Pi really makes that a pain. It is conceptually too software centric. Python has become a joke.

hzeller:
Talk is cheap. Show me the code.
If you want to contribute, send pull requests, David.

PR #1182, opened 2020-10-23
Point to https://rpi-rgb-led-matrix.discourse.group/ for help

marcmerlin:
Point to https://rpi-rgb-led-matrix.discourse.group/ for help


marcmerlin:
HI @hzeller  I can tell that you're not really working on the code right now (which is fine), but would you mind adding this pointer so that people can start helping one another on the discord group?
Thanks

hzeller:
Thanks!

PR #1195, opened 2020-11-10
Fix skipping of large number of frames

amund7:
An older version did this OK, but the current would never seem to finish skipping and start playing (when skipping 20.000 frames).


marcmerlin:
Apologies for the lack of answer.  Is this still needed? If so, can you send an update against the latest branch (it doesn't apply cleanly) and I'll merge it.

marcmerlin:

PR #1253, opened 2021-02-15
Introduce --led-no-drop-privs for Python samples

jonico:
* introduce `--led-no-drop-privs` to *samplebase.py* so that it is possible to keep root permissions in the python samples if needed


hzeller:
Thanks! merged.

PR #1257, opened 2021-02-19
Add meson build system

mostthingsweb:
This makes it easier to build under Yocto.

Components that are built: librgbmatrix, the example applications, and the Python extensions.


mostthingsweb:
Now that I think about it I shouldn't have deleted the generated .cpp files because otherwise users of the Makefile can't use this. So I'll update the pull request to fix it.

marcmerlin:

I will however link to the master python bug so that if reviewers feel this is helpful, it can be re-opened and pulled in.

Issue #1266, opened 2021-03-05
samplebase.py does not accept '--led-gpio-mapping=regular-pi1'


beta-tester:
hi,

i just dug up my old Raspberry Pi 1B  rev.1 and let it drive my also old led-matrix panels.
first i stubled around with the RPi 1B rev1 related issue of the missing green color at the top half of the matrix.

after i found the solution to that issue by using `--led-gpio-mapping=regular-pi1`, i now observed another issue.

`rpi-rgb-led-matrix/bindings/python/samples/samplebase.py` does not accept that option.

```
error: argument -m/--led-gpio-mapping: invalid choice: 'regular-pi1' (choose from 'regular', 'adafruit-hat', 'adafruit-hat-pwm')
```

i can add that option to the list of accepted choices in `samplebase.py` but i guess that there are other options/choices missing for other parameters. maybe this file is not up to date.

hzeller:
Yes, I think old Pi's are so rare and using Python on a Pi1 is very slow to use that nobody encountered this omission yet.
Please send a pull request with that choice added.

PR #1267, opened 2021-03-05
added 'regular-pi1' to --led-gpio-mapping choices

beta-tester:
see: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1266


hzeller:
Thanks, merged!

PR #1268, opened 2021-03-06
added more missing values to --led-gpio-mapping choices

beta-tester:


hzeller:
I'd not add 'classic' or  'classic-pi1'. Nobody should be using them.

beta-tester:
ok, no problem, then i close that request.

PR #1357, opened 2021-10-30
TTF to BDF converter

K0rkunc:
You can convert all fonts installed on your Windows computer to BDF fonts of any size.


marcmerlin:
Please don't take this the wrong way, but it's not safe for us to include unknown windows binaries to this project.
If you want to host your converter somewhere and provide a link to it, I'd be willing to include that, but honestly it would be nicer and safer if you can provide source for that converter so people know it's safe.
I'm closing this but you can reply to it and re-open as needed.

PR #1377, opened 2021-12-10
Add support for active high output enable

maximus64:
This change added support for active high output signal for OE and also bring in support for my custom LED matrix cube hat name `cubie-hat`. The output enable pin on this hat is inverted to avoid glitch on initial power on. 

Background:
The 3.3v power rail on Raspberry Pi 4 comes on later than the 5v input supply and since the OE is pulled up to the 3.3v rail on the active3 hat, there will be a short glitch on the OE signal at powering up causing unwanted flicker on the LED panel. To fix this issue, I added an NPN transistor to drive the OE signal on my custom hat and that fixed the issue.
image

### DIFF
--- a/lib/framebuffer.cc
+++ b/lib/framebuffer.cc
@@ -460,6 +460,7 @@ Framebuffer::~Framebuffer() {
     if (b >= dither_bits) timing_ns *= 2;
   }
   sOutputEnablePulser = PinPulser::Create(io, h.output_enable,
+                                          h.output_polarity,
                                           allow_hardware_pulsing,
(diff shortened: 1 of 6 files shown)


marcmerlin:
this looks interesting but only seems to be for you.
Can you convince a PI Hat maker to add this hardware? If you do, I'll be happy to merge this (although please remove all diffs that are whitespace only and do not include code changes).
I am cleaning up the big backlog of requests, so for now I will close this until you have an update, however do feel free to re-open with an updated PR and link to a board that will support this new code.

PR #1388, opened 2022-01-09
Add new test & new panel

rotrico:
Hi Mr. Henner,
First,  my apologies if this is isn't the way to perform a pull request, I'm not sure how it works github.

I added support for a new panel that It wasn't supported by your library or at least I wasn't able to get it work correctly in any of 19 multiplex modes, addr type, etc,. It's a panel sold by a chinese manufacturer, probably other panels in market use same logic to address pixels, I think that it can be useful for other people. https://www.led-card.com/outdoor-p5-smd-fullcolor-module.html

Also I added a new test (12)  in demo program to move and change primary color of one pixel for all of screen, it was useful for me to discover the address logic behind my led panel.

Regards,


marcmerlin:
please accept our apologies for not getting back to you earlier, it's only volunteer time and there wasn't enough of it.
Can you check the current mappers to see if they fixed what you needed, and if not can you send an updated PR that applies cleanly against the top of tree? (head)
To make space, I'm resolving this until you have an update but do feel free to reopen/followup

PR #1443, opened 2022-06-30
update Qiangli Q10 link

AsebWebDev:
The link provided in the readme is out of date (404), so I switched to the correct one (hopefully). Not sure if X10 is an alias for the same product, but I could not find a product listed, so I assume it is not on the market anymore. Hope this helps.


marcmerlin:
they both seem 404 :-/  can you update with a link working today?

marcmerlin:

PR #1447, opened 2022-07-19
Add framerate_fraction to RGBMatrix.SwapOnVSync() python binding

SolidHal:
Included a sample which uses it to efficiently display gifs at a
reasonable speed


hzeller (approved):
Very cool. Thanks for your contribution!

PR #1457, opened 2022-08-22
Python bindings for FrameCanvas::Serialize/Deserialize

SolidHal:
this enables usage like in cpp content-streamer

### DIFF
--- a/bindings/python/rgbmatrix/core.pyx
+++ b/bindings/python/rgbmatrix/core.pyx
@@ -53,6 +53,12 @@ cdef class Canvas:
                 b = (pixel >> 16) & 0xFF
                 my_canvas.SetPixel(xstart+col, ystart+row, r, g, b)
 
+cdef class FrameData:
+    cdef const char *__getData(self) except +:
+        if <void*>self.__data != NULL:
+            return self.__data
+        raise Exception("FrameData ptr was destroyed or not initialized, you cannot use this object anymore")
+
 cdef class FrameCanvas(Canvas):
     def __dealloc__(self):
         if <void*>self.__canvas != NULL:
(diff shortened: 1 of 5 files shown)


bindings/python/rgbmatrix/core.pyx
```diff
@@ -53,6 +53,12 @@ cdef class Canvas:
                 b = (pixel >> 16) & 0xFF
                 my_canvas.SetPixel(xstart+col, ystart+row, r, g, b)
 
+cdef class FrameData:
+    cdef const char *__getData(self) except +:
```
ledvinap, line 57:
```suggestion
    cdef const char *__getData(self) except NULL:
```
`except +` means that code can throw C++ exception
`except NULL` - check exception state if returned value is NULL

bindings/python/rgbmatrix/core.pyx
```diff
@@ -72,6 +78,17 @@ cdef class FrameCanvas(Canvas):
     def SetPixel(self, int x, int y, uint8_t red, uint8_t green, uint8_t blue):
         (<cppinc.FrameCanvas*>self.__getCanvas()).SetPixel(x, y, red, green, blue)
 
+    def Serialize(self):
+        cdef const char* data
+        cdef size_t length
+        (<cppinc.FrameCanvas*>self.__getCanvas()).Serialize(&data, &length)
```
ledvinap, line 84: I'm afraid that this won't work for your purpose. Serialize just returns pointer to internal frame representation (one instance per FrameCanvas), it does not copy it.

One option is to implement buffer protocol for `FrameCanvas` and then use python for memory management. I'll post PR soon

marcmerlin:
Per @ledvinap  I'm going to assume we can't merge this.

PR #1477, opened 2022-11-01
Python refactor

ledvinap:
(For discussion)

- Refactor python bindings a bit, fixing various problems leading to SEGFAULT
- class passed to SwapOnVSync is returned on second call unmodified
- FrameCanvas is recycled when python object is freed

Add support for FrameCanvas::Serialize:
- Python class with Buffer Protocol is returned
- It can be easily used to write to file etc. (and no copy is necessary)
- Now read-only, but read-write should be safe and it will allow zero-copy writing to framebuffer
  - another option is to export Deserialize, possibly with Buffer Protocol interface, but performance will be slightly smaller


marcmerlin:
is this still needed after @chrisgilldc and @ty-porter 's changes?

marcmerlin:
@ledvinap  please look at https://github.com/hzeller/rpi-rgb-led-matrix/pull/1753
I'd like to get all the python updates/fixes in sync and get the best merge of everything.

ty-porter:
This looks great, supporting canvas serialization would be a welcome addition to the Python bindings! This PR might need to be rebased, looks like it is touching the same code as #1817 but I believe it should be compatible.

bindings/python/rgbmatrix/core.pyx
```diff
@@ -193,45 +274,64 @@ cdef class RGBMatrixOptions:
...
-        self.__matrix.Fill(red, green, blue)
+    def __dealloc__(self):
+        if self.__matrix != NULL:
+            self.__matrix.Clear()
+            del self.__matrix
 
-    def SetPixel(self, int x, int y, uint8_t red, uint8_t green, uint8_t blue):
-        self.__matrix.SetPixel(x, y, red, green, blue)
+    cdef cppinc.Canvas* __getCanvas(self) except NULL:
+        assert self.__matrix != NULL
```
ty-porter: Generally: Are these assertions required? If so, can we at least add an error message?

marcmerlin:
yeah, this needs a rebase now, but if you're good with it on principle, happy to merge after that (please test the full code after rebase to make sure nothing broke)

marcmerlin:
@ledvinap  and @ty-porter  I'm not able to tag or add you to the master bug to keep track of python stuff. Can you add yourselves as assignees to https://github.com/hzeller/rpi-rgb-led-matrix/issues/1749 or just reply to the issue, and then I'll probably be able to add you so you have the option to review and vet new python changes?

marcmerlin:

chrisgilldc:
I'd defer to @ty-porter, I don't think my changes would have intersected here and I don't have Cython expertise.

marcmerlin:
@ledvinap I'm not rejecting your idea or CL in general, but this has been open for too long without progress and won't mere as is.
Please submit a new CL .  @ty-porter  given that you liked this, it's also ok for you to make a new version since we haven't heard form @ledvinap  in a while

PR #1478, opened 2022-11-01
Feature remap mapper

ledvinap:
(for discussion, needs some polishing, needs testing)

This PR implements universal mapper - each segment is mapped to arbitrary position and orientation on canvas

My use case is two separate displays connected as chains, each with different geometry. But it shall handle almost all possible panel configuration

Syntax is:
```
--led-pixel-mapper='Remap:<new_width>,<new_height>|<panel0_x>,<panel0_y><panel0_orientation>|<panel1_x>,<panel1_y><panel1_orientation>'
```
new_width, new_height - size of created canvas. It can be both larger and smaller than old canvas.
panel0_x, panel0_y - upper-left corner of remapped panel
panel0_orientation - `n`,`s`,`e`,`w` for panel orientation, `x` to discard panel

Exactly one entry must be specified for each LED panel (chain * parallel entries). 
Mapping may be partially outside new canvas (maybe useful for something?)
Unused positions must be discarded (`0,0x`) - useful if chains are of different length
It is possible let some canvas space unused (no panel mapped to it). Writes to this area will be ignored.

For example:
```
 --led-cols=16 --led-rows=8 --led-chain=5 --led-parallel=2 --led-pixel-mapper='Remap:40,32|0,0e|8,0e|16,0e|24,0e|32,0e|0,16n|16,16n|0,24n|16,24n|0,0x'
```
First chain is 5 panels in line, top of the panel is pointing right, 40x16 pixels
Second chain is 2x2 square, panels pointing up, 32x16 pixels, top-left is 0,16 in canvas
Last panel on second chain is not used
```
1 2 3 4 5
1 2 3 4 5
6 6 7 7 
8 8 9 9
```

I'll add some documentation if there is interest in this feature

### DIFF
--- a/include/pixel-mapper.h
+++ b/include/pixel-mapper.h
@@ -80,6 +80,14 @@ public:
   virtual void MapVisibleToMatrix(int matrix_width, int matrix_height,
                                   int visible_x, int visible_y,
                                   int *matrix_x, int *matrix_y) const = 0;
+
+  virtual bool MapMatrixToVisible(int matrix_width, int matrix_height,
+                                  int matrix_x, int matrix_y,
+                                  int *visible_x, int *visible_y) const { return false; };
+  enum MappingType {
+    VisibleToMatrix, MatrixToVisible
+  };
+  virtual MappingType GetMappingType() const { return VisibleToMatrix; }
 };
 
 // This is a place to register PixelMappers globally. If you register your
--- a/lib/led-matrix.cc
+++ b/lib/led-matrix.cc
@@ -558,6 +558,8 @@ bool RGBMatrix::Impl::ApplyPixelMapper(const PixelMapper *mapper) {
   }
   PixelDesignatorMap *new_mapper = new PixelDesignatorMap(
     new_width, new_height, shared_pixel_mapper_->GetFillColorBits());
+  switch (mapper->GetMappingType()) {
+    case PixelMapper::VisibleToMatrix:
   for (int y = 0; y < new_height; ++y) {
     for (int x = 0; x < new_width; ++x) {
       int orig_x = -1, orig_y = -1;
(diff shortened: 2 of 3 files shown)


ledvinap:
Fixed panel ordering, panels are now counted from Raspberry (as in example above)

iamlegio:
dear ledvinap,

I got the mapper to register and now I am in the process of understanding how to map it properly:

I have 6x8 modules, standing 16 width 32 height. So that would be like this?
MatrixRemap

Thank you for your help!
Thomas

ledvinap:
Yes, that's how it should work. You need to map all panels (end of chain 3), so you must discard last 6 panels
```
 --led-pixel-mapper='Remap:192,128|0,0s|0,32s|...|112,32n|112,0n|0,0x|0,0x|0,0x|0,0x|0,0x|0,0x'

iamlegio:
thank you, yes I discarded the last panels, was a little sloppy and did not write it out :)

0,0 is in top right Corner but is it seen from the front or the back? 

You wrote ‘Remap:192,128’ should it not be 128 width and 192 height. Remap:<new_width>,<new_height>| Or should I flip it with ;Rotate after the mapping?

My screen is built to have 0,0 in lower left corner seen from the back. So it is a little confusing to map it correctly, hence all the questions.

ledvinap:
Sorry, was answering too late in the evening yesterday.

The mapping uses rgbmatrix coordinates - top-left corner is (0,0), x increasing to the right, y is increasing down, looking on leds (front side). Display coordinates are used in mapping, panel (and chain) is implicit - |<chain1, panel 1>|<chain1, panel2|...|<chain 2, panel 1>|... 

Size will be `128,192`

If I understand it correctly, first panel on chain 1 in your case in bottom right corner when looking on leds? In that case, first entry would be `|112,160n|`

iamlegio:
It Works!

my panels are oriented like this:

PanelOrientation

So instead of north and south I had to use east and west:

Remap:192,128|112,160e|...|0,160w|0,0x|0,0x|0,0x|0,0x|0,0x|0,0x

Thank you so much for your help and excellent work in writing this extension!
Thomas

ledvinap:
Great!

Direction should be where top of module  (4 arrows on PCB) are pointing.

BTW: It is possible to rearrange panels so that all chains are 16 panels (you will gain ~11% refresh increase). 
And you get shorter cables from PI as bonus ;-)

RGB

iamlegio:
Again Thank You,

yes I could defiantly rearrange the modules. Will try it now, great idea!

AlexRoidl:
+1

hzeller:
Nice! This is the kind of universal mapper that I was looking forward for someone to implement!

Sorry for the delay in the review, I was busy around November and might not have noticed the pull request.

I still have to look at the details of the code, and will probably have time on Sunday for that.

In the meantime: to be most useful to users, can you add a description in the mapper section of the README in examples-api-use/ ?
https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/examples-api-use#remapping-coordinates

ledvinap:
@hzeller : 
- Can you look at parametr syntax? It is still time to improve it, it will be more complicated later.
- I'll fix some minor problems before merge (typos + whtespace)
- I hate writing documentation ;-). But I'll try, most of it is in this PR anyway

hzeller:
How is the working coming along with the documentation ?

AlexRoidl:
@hzeller Is it possible to use this in the python binding somehow? Even if it is only a hot-fix?

marcmerlin:
there is still demand for this for sure. @ledvinap  what's the current status? I'm happy to help review and merge this.

marcmerlin:
Ok, more PRs than I wish got dropped because they wouldn't merge anymore and this one does.
I'm going to merge it because a non documented untested mapper that doesn't affect other things is better than none (this is mostly a stance from Linus on Linux and new drivers).
@ledvinap  please do test this as merged and please provide some basic documentation as followup PR which I will approve on the spot.
If this works as intended, it will fix most mapping needs I believe, which will be awesome.

marcmerlin:
submitted your docs in 
https://github.com/hzeller/rpi-rgb-led-matrix/commit/1414585dc1328892356ee5e6c2690ec4b3986381

PR #1479, opened 2022-11-08
Pi4 gpio timing experiment 

ledvinap:
(for discussion, do not merge)

This PR uses DSB memory barrier that forces AXI transaction to complete before next instruction. 

Pros:
- You get very accurate timing, RGB data and clock are almost perfectly synchronized. Absolutely no ghosting
- possibly slightly lower power consumption (5% ? core is stalled, not busy waiting)

Cons:
- Clock frequency si limited, to about 6MHz on non-overclocked PI4 
  - about half vs using --led-gpio-slowdown. 
  - Overclocking does change timing, but I did not investigate it yet. 
-  Memory access from other threads will slow serial speed down a bit (but timing is not degraded as in --led-gpio-slowdown case)

It may be very useful with long/noisy cables or slow matrices. Timing margins are much beter

dsb-st-clock

### DIFF
--- a/examples-api-use/minimal-example.cc
+++ b/examples-api-use/minimal-example.cc
@@ -33,7 +33,7 @@ static void DrawOnCanvas(Canvas *canvas) {
   float angle_step = 1.0 / 360;
   for (float a = 0, r = 0; r < radius_max; a += angle_step, r += angle_step) {
     if (interrupt_received)
-      return;
+    return;
     float dot_x = cos(a * 2 * M_PI) * r;
     float dot_y = sin(a * 2 * M_PI) * r;
     canvas->SetPixel(center_x + dot_x, center_y + dot_y,
--- a/lib/gpio.h
+++ b/lib/gpio.h
@@ -47,26 +47,23 @@ public:
   inline void SetBits(gpio_bits_t value) {
     if (!value) return;
     WriteSetBits(value);
-    for (int i = 0; i < slowdown_; ++i) {
-      WriteSetBits(value);
-    }
+    delay();
   }
 
   // Clear the bits that are '1' in the output. Leave the rest untouched.
   inline void ClearBits(gpio_bits_t value) {
     if (!value) return;
     WriteClrBits(value);
-    for (int i = 0; i < slowdown_; ++i) {
-      WriteClrBits(value);
-    }
+    delay();
   }
 
   // Write all the bits of "value" mentioned in "mask". Leave the rest untouched.
   inline void WriteMaskedBits(gpio_bits_t value, gpio_bits_t mask) {
     // Writing a word is two operations. The IO is actually pretty slow, so
     // this should probably  be unnoticable.
-    ClearBits(~value & mask);
-    SetBits(value & mask);
+    WriteClrBits(~value & mask);
+    WriteSetBits(value & mask);
+    delay();
   }
 
   inline gpio_bits_t Read() const { return ReadRegisters() & input_bits_; }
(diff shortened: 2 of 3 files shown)


ledvinap:
@hzeller : BTW: Did you consider/invvestigate using videocore for data generation? It may be possible with great performance ...

ledvinap:
I've changed PR a bit. Now either dummy writes or DSB is used for slowdown.

Appending dummy writes works better than old code - register value is updated first, then delay is inserted.
Old code did: clear bits, delay, set bits, delay (set clock, delay) - wasting half of delay time after pushing column data to IO ports. This should result in about 30% increase of clock speed with no negative side effects.

DSB is used with  `--led-gpio-slowdown=-1`. Code must be compiled for ARM7 or ARM8, `-march=native` handles this now.

It may be better to emit 'DSB ST' opcode directly (so assembler won't complain) and test for supported CPU on runtime.

hzeller:
Ah, sorry for the delay. Sometimes the messages by github just go under in a flood of other mails.

I'll have a closer look at this, it looks like you have more insights into how the internal GPIO hardware subsystem working. My delay is just something that was based on my early experiments with the first Pi 1 12 years ago, so revisiting how GPIO is paced is overdue.

I am currently travelling, and will have do some Oscilloscope measurements on all Pi versions I have access to when I come back end of August before merging.

Thanks for your contribution and sorry that I took so long to see it.

(yes I did investeigate the videocore for nicely timed writes, and I have a half-written patch in one of my local branches somewhere, but I did not have much time yet to finish it)

examples-api-use/minimal-example.cc
```diff
@@ -33,7 +33,7 @@ static void DrawOnCanvas(Canvas *canvas) {
   float angle_step = 1.0 / 360;
   for (float a = 0, r = 0; r < radius_max; a += angle_step, r += angle_step) {
     if (interrupt_received)
-      return;
+    return;
```
hzeller, line 36: did you intent do remove the indentation here ?

hzeller (approved):

examples-api-use/minimal-example.cc
hzeller: (I'll fix that separately)

hzeller:
Thanks, merged; since this is only enabled with slowdown -1, the regular behavior should be the same (execpt the single-delay for writing the masked bits, but that is benign and might need some different slowdown experiments by the user)

bluelasers:
In summary, I think -1 is a nice compromise. There are no winners on GPIO slowdown and most of the time we never care.

Officially supported:
- 16x32 (11.1MHz-22.2MHz)
- 32x32 (11.1MHz-22.2MHz)
- 32x64 (5.5MHz-11.1MHz)
- 64x64 (5.5MHz-11.1MHz)

Not officially supported:
- 64x128 (2.7MHz-5.5MHz)
- 64x256 (1.3MHz-2.7MHz)
- 64x512 (0.6MHz-1.3MHz)

These are based on RC time constants: (t = 5RC, f = 1/t)
- 10 time periods (5 per half)
- 50 ohms
- 15 pF
- 6 loads per 16 columns (outdoor panels are higher)

Software may not know the overclock or the network congestion. Technically those problems are not a concern for this project when used as a dedicated IO device. (6MHz with BCM also prevents another configuration issue.)

marcmerlin:
Eh @bluelasers , es-tu Dave Tatcher (or whatever the spelling was)?
regardless of the answer, I see you have a new account, contributed nothing to the source code or any other project, and if it's going to be another string of random drive by comments with 0 actual code contributions of any kind, you're going to be invited, again, to go spend your time elsewhere.

PR #1518, opened 2023-02-24
Change the members position to have the same order in c and c++

mick111:
To solve issue #1517 about order of members of structures RGBMatrix:Options and RGBLedMatrixOptions


hzeller:
Thanks, somehow this got messed up. Thanks for the fix.
One request though: can you move the c-struct members around instead of the c++ members ?
I want to keep ABI compatibility in the c++ lib.

hzeller:
Thanks, merged!

PR #1528, opened 2023-03-29
Add a link to the V2 version of the Active-3 board.

emilytocin:
Sorry this doesn't have full details. I'm unsure the V2 implementation is REALLY the full implementation of the Active3 board, but it's the one I FAR prefer for all my projects, and the one I always am looking for when I come to this page.


Ha2k4r (approved):
This is perfect and something this library needs for an update we need to merge this with the main branch

marcmerlin:
I added this independently in https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/adapter/README.md
apologies we didn't get back to you earlier.

PR #1534, opened 2023-04-18
Respect BBX offset in glyph parser

### LINKED ISSUE #1521: DrawText() always left-aligns glyphs even when they specify an offset in BBX field
Hi! I maintain [mlb-led-scoreboard](https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard) which leverages this library.

We have had an issue reported where users noticed some strange alignment issues for some characters.
https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/issues/416#issuecomment-1453977463

Our default font is `tom-thumb.bdf`. Any glyphs which specify an X offset in their `BBX` field will be left-aligned instead of correctly aligned by the bounding box. An example character is `!`, where `BBX 1 5 1 0` uses the third column to set a X-offset of +1.
https://github.com/hzeller/rpi-rgb-led-matrix/blob/0bbf64d3c145e65ced8121befc1041c15f38c4bc/fonts/tom-thumb.bdf#L41-L52

Picture of some other characters (`!`, `i`, `:`, `.` are affected, `x` for reference spacing):
Example

I was able to trace it down to https://github.com/hzeller/rpi-rgb-led-matrix/commit/39953b7a6e5e0ae6d4ff0a7aba0f9b234b878d9b. Our previous advice to end users was to revert back to https://github.com/hzeller/rpi-rgb-led-matrix/commit/a56338db0f003d5236f2ce98c73a591d64a70852, but that commit is over a year old at this point.

WardBrian:
This would close #1521

The commit identified there (https://github.com/hzeller/rpi-rgb-led-matrix/commit/39953b7a6e5e0ae6d4ff0a7aba0f9b234b878d9b) changed the font parsing to no longer need a left shift, but this meant the offset was no longer used at all. 

This introduces a right shift. On the font I was using (https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard/blob/2bbabaf6c100bd5df197abf76943d08d8c39f3aa/assets/fonts/patched/4x6.bdf) this now re-recreates the exact parse as the older code (except it is now 196 bits wide, rather than 64 bits), and the font characters are now correctly offset.


WardBrian:
Gentle ping @hzeller - this one liner fixes and issue affecting https://github.com/MLB-LED-Scoreboard/mlb-led-scoreboard

hzeller:
sorry, just now seeing this. Sometimes github notifications drown between other stuff.

PR #1556, opened 2023-06-19
Re-use CreateFrameCanvas() in samples/gif-viewer.py

redphx:
I used the original sample to display a list of GIF files.  
After a few minutes I got this warning:  
```
CreateFrameCanvas() called 500 times. Usually you only want to call it once (or at most a few times) for double-buffering...
```
I modified the sample and re-use the canvas instead.


marcmerlin:
thanks for submitting this fix. Can you update against master and I'll merget this.
If it's not easy for you to do this with git, you can also make a new PR and I'll merge that instead (please add a pointer to this one in that case so I can delete it)

redphx:
@marcmerlin Updated https://github.com/hzeller/rpi-rgb-led-matrix/pull/1827

PR #1561, opened 2023-07-02
examples: Update .gitignore

somechris:
`input-example`, and `pixel-mover` are built files, so we ignore them.


neuschaefer:
looks good to me

PR #1563, opened 2023-07-02
Allow to sleep instead of busy waiting when limiting refresh rate

somechris:
On single core boards, busy waiting renders the system unresponsive. We add the --led-no-busy-waiting flag, to allow the process to sleep instead of busy waiting, thereby freeing up CPU cycles for other processes at the cost of slightly less accurate frame timing.

On a Raspberry Pi 3B on two 32x32 panels, busy waiting keeps a core 98% busy. Sleeping reduces it to 14%.

On a Raspberry Zero on two 32x32 panels, busy waiting keeps the single core 95% busy and gives an unresponsive system. Sleeping reduces it to 32% and keeps the system responsive.

Sleeping degrades the timing accuracy a bit, but is still good enough by a far margin for many uses. The following numbers are for driving two 32x32 panels and averaging the refresh rate across a minute (the 1kHz refresh rate limit is of course not reached, and is only there to make sure that the relevant code path is taken):

                                                   +--------------+-----------+
                                                   | busy waiting | sleeping  |
+--------------------------------------------------+--------------+-----------+
| Raspberry Pi 3B, limit 70Hz, hardw. pulsing      |  70.000Hz    |  69.977Hz |
| Raspberry Pi 3B, limit 70Hz, no hardw. pulsing   |  70.000Hz    |  69.977Hz |
| Raspberry Pi 3B, limit 1kHz, hardw. pulsing      | 430.824Hz    | 430.855Hz |
| Raspberry Pi 3B, limit 1kHz, no hardw. pulsing   | 365.498Hz    | 365.456Hz |
| Raspberry Pi Zero, limit 70Hz, hardw. pulsing    |  69.986Hz    |  69.643Hz |
| Raspberry Pi Zero, limit 70Hz, no hardw. pulsing |  69.986Hz    |  69.638Hz |
| Raspberry Pi Zero, limit 1kHz, hardw. pulsing    | 263.877Hz    | 264.538Hz |
| Raspberry Pi Zero, limit 1kHz, no hardw. pulsing | 233.166Hz    | 231.650Hz |
+--------------------------------------------------+--------------+-----------+

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,7 +445,7 @@ to debug if it has something to do with the sound subsystem (see Troubleshooting
 section). This is really only recommended for debugging; typically you actually
 want the hardware pulses as it results in a much more stable picture.
 
-<a name="no-drop-priv"/>
+<a name="no-drop-priv"></a>
 
 ```
 --led-no-drop-privs       : Don't drop privileges from 'root' after initializing the hardware.
--- a/README.md
+++ b/README.md
@@ -381,6 +381,20 @@ the vsync-multiple flag `-V` in the [led-image-viewer] or
 [video-viewer] utility programs.
 
 ```
+--led-no-busy-waiting     : Don't use busy waiting when limiting refresh rate.
+```
+
+This allows to switch from busy waiting to sleep waiting when limiting the
+refresh rate (`--led-limit-refresh`).
+
+By default, refresh rate limiting uses busy waiting, which is CPU intensive but
+gives most accurate timings. This is fine for multi-core boards.
+
+On single core boards (e.g.: Raspberry Pi Zero) busy waiting makes the system
+unresponsive for other/background tasks. There, sleep waiting improves the
+system's responsiveness at the cost of slightly less accurate timings.
+
+```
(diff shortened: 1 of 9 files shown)


wawanbreton:
Hi @somechris, thanks for this improvement, this really helps lowering the CPU load, very useful on a Pi Zero :)
I would even suggest adding it without an option, I fail to see a case where you would actually want to do an active wait. Maybe the timing is a bit less perfect indeed, but as long as you have a decent refresh rate, the display is as good !

hzeller:
Nice! Thanks @somechris 
Sorry, just seeing this now as @wawanbreton 's mail bumped the thread. The notifications on github are easy to miss.

The code looks great. Thanks for taking care that this is an inverted option so that memsetting the struct to zero would be backward compatible.

So code-wise, I think we're ready to merge, but you should also add the description of this flag in a paragraph in the README where `--led-limit-refresh` is explained. If it is in the README, it is easier discoverable. Can you update the PR ?

somechris:
> Can you update the PR ?

Done :-)

PR #1564, opened 2023-07-02
bindings/python: Add priv_drop_user and priv_drop_group

somechris:
This gives Python users control over which user/group the process is dropped into.


hzeller:
Sorry for the delay, just going through old PRs :)
Could you re-base as the Python bindings have been touched recently.

somechris:
Will do, but it'll take me a few days to get to it.

hzeller:
The generated files are not included anymore to avoid staleness so only changes in the *.pyx need to be submitted. So I suspect it will be fairly straigthforward to update the patch.

somechris:
> Could you re-base as the Python bindings have been touched recently.

Done.

Thanks for removing the `core.cpp`. It's so much simpler without it! \o/

hzeller:
Thanks, merged!

PR #1584, opened 2023-08-23
fix: auto `b` variable in loop is shadowning `b` fcn argument

rochaferraz:
1. Minor fix that changes a variable name. Some compilers options (such as what I'm using) don't like that.
2. Explicit type casting on a few led-matrix.cc methods


hzeller:
Thanks!

PR #1586, opened 2023-08-24
Added additional Troubleshooting point

Serzhs:


hzeller:
So was this something that didn't work on a say Raspberry Pi 3 but worked on a Pi 4 ?

When reading that you connected a sensor it sounds like you might've used GPIO pins that were used in by the panel (and which you then would use to route to pins in the hardware-mapping.c)

Serzhs:
@hzeller I initially used a super old Raspberry Pi 1, and it didn’t work. However, when I tried the same thing on a Raspberry Pi 3, it worked automatically.

marcmerlin:
can't hurt to add even if I doubt it will help many :)

PR #1590, opened 2023-09-08
Updated ZStripeMultiplexMapper to support variable tile sizes

andytheengineerguy:
This change allowed me to use the ZStripeMultiplexMapper with a tile height of 13, for an LED panel that required this change for the mapping to work

I decided to update ZStripeMultiplexMapper - rather than make a new mapping - because this change may be helpful for future panels that require different tile heights or widths, like mine.

I've updated the existing mappings so that their original tile size values (8 and 4) are accounted for.

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -153,14 +153,16 @@ public:
 
 class ZStripeMultiplexMapper : public MultiplexMapperBase {
 public:
-  ZStripeMultiplexMapper(const char *name, int even_vblock_offset, int odd_vblock_offset)
+  ZStripeMultiplexMapper(const char *name, int even_vblock_offset, int odd_vblock_offset, int width, int height)
   : MultiplexMapperBase(name, 2),
     even_vblock_offset_(even_vblock_offset),
-    odd_vblock_offset_(odd_vblock_offset) {}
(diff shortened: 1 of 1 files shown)


lib/multiplex-mappers.cc
```diff
@@ -153,14 +153,16 @@ class SpiralMultiplexMapper : public MultiplexMapperBase {
...
     even_vblock_offset_(even_vblock_offset),
-    odd_vblock_offset_(odd_vblock_offset) {}
+    odd_vblock_offset_(odd_vblock_offset),
+    width_(width),
+    height_(height) {}
 
   void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
-    static const int tile_width = 8;
-    static const int tile_height = 4;
+    static const int tile_width = width_;
```
hzeller, line 164: if you use width_ and height_ member variables, you can't assign this to static variables here, as they are global and override them with whoever ran last.

Just replace all uses of `tile_height`, and `tile_width` with the new members `height_` and `width_`.

hzeller (changes requested):
Looks good in general, but make sure to not modify global variables messing with other instances of the same class.

marcmerlin:
if you have an updated version of this, please update this issue. for now, I'll resolve it until you have an update.

PR #1612, opened 2023-12-27
Expose drop_priv_user and drop_priv_group to python interface

mxklabs:
I needed to be able to drop back to a specific user/group in my Python application and found that while these options were available in the C++ interface they were not exposed to Python. 

This PR adds bindings to allow Python users to drop back to a specific user/group.


somechris:
This seems to be a dupe of #1564

hzeller:
Yes, looks like a duplicate. Let's get #1564 ready to merge.

PR #1617, opened 2024-01-08
Add Java bindings

nullEuro:
This pull request introduces a Gradle project to implement a JNI bridge for the C++ APIs. This addition allows Java applications to interact with LED matrices. 

Tbh I will probably be too lazy to maintain this in the future if any bigger breaking changes arise. But I do use the project myself to drive a matrix in my living room.


nullEuro:
Completes #1491

hzeller:
This looks like a pretty involved binding for which it will probably be hard for the maintainers to keep up-to-date, in particular since you already expressed that you don't consider to maintain it.

I encourage to host this as a separate project.

Issue #1626, opened 2024-02-28
[Q&A] High priority thread restricted by low priority thread?


bluelasers:
I was looking into the code for possible points of concern. To me this looks like a possible deadlock which results in a stall of the high priority thread. The front ground thread is capable of acquiring the lock and losing the CPU(s). What happens to the background thread during this?
https://github.com/hzeller/rpi-rgb-led-matrix/blob/a3eea997a9254b83ab2de97ae80d83588f696387/lib/led-matrix.cc#L224-L231

https://github.com/hzeller/rpi-rgb-led-matrix/blob/a3eea997a9254b83ab2de97ae80d83588f696387/lib/led-matrix.cc#L173-L187

It would seem to me that you would try the lock and pass it over if held. Can you create critical section in user space?

hzeller:
Can you explain where you see a deadlock ? A deadlock is if someone is blocking a mutex `a` and waits for a mutex `b`; and the someone else is blocking mutex `b` but waits for mutex `a` to succeed.

We don't have this situation here: just one mutex used in two critical sections that should be run exclusively. Yes you can temporarily use CPU in one of these, elongating the time the other has to wait, but you never stall.

Also note that each of the sections is short, does not do any external IO and just swaps around variables.

bluelasers:
OS cannot be predicted. If you lose the CPU you lose the background thread. Basically this means the multicore is single threaded. The cache can get cold during this time. Depending on the version of hardware this may matter more or less.

You are SMT program currently and I am trying to get you back to where you tried to be. I would recommend removing all but core zero from the scheduler. I will just leave you with the knowledge of that option as I am not sure you will agree with it.

By the way there is another way to write this which is more SMT. This works around this issue completely, but I have not written a full version of it. So I just leave you with the knowledge of its existence only, as it would be a good size rewrite.

hzeller:
What you describe is essentially priority inversion, and yes, it can be problematic if there is only one core and a saturated realtime thread.

And yes, this will always be a theoretical issue (but not in practice).

bluelasers:
My concern is outlined in #1468. This issue applies on multicore also. We can mask out the background threads priority. This stalls the panels multiplexing. We may be forced to set out a full tick randomly. That is a relatively long period of time for multiplexing. Certain workloads may be more prone to this issue. (Multithreading for example, see #1678.)

[...]

PR #1659, opened 2024-05-29
#1626 Use try lock method to avoid stalling real time IO thread.

bluelasers:
See issue for details.

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -61,6 +61,7 @@ public:
   Mutex() { pthread_mutex_init(&mutex_, NULL); }
   ~Mutex() { pthread_mutex_destroy(&mutex_); }
   void Lock() { pthread_mutex_lock(&mutex_); }
+  bool try_lock() { return pthread_mutex_trylock(&mutex_) == 0; }
   void Unlock() { pthread_mutex_unlock(&mutex_); }
 
   // Wait on condition. If "timeout_ms" is < 0, it waits forever, otherwise
--- a/lib/led-matrix.cc
+++ b/lib/led-matrix.cc
@@ -170,8 +170,7 @@ public:
         ->DumpToMatrix(io_, start_bit_[low_bit_sequence % 4]);
 
       // SwapOnVSync() exchange.
-      {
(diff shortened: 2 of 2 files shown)


hzeller:
Do you ever experience that situation or is this the problem #1626 you think you could solve this way ?

The problem with try-lock is, that you now might never give the user a chance to update the frame as it will allow the update thread to hog the CPU solidly (as it is high-prio and if you have a single core that might starve the regular priority threads). With the actual lock you will give a chance to briefly give up that core.

bluelasers:
This pull request is the handcuff to the issue. It is part of a process I try to follow.

I am under the impression that mutex grab is blocking. The condition wait will release the mutex from blocking. Try lock will attempt a grab is it can, but will never block. If it gets the lock it will do the condition signal, otherwise skip over it on this pass. This means the high priority thread can never block.

There is a risk of locking out the low priority thread. It is the exact same problem just in reverse. Perhaps a pipe would be better? This allows testing the ends of the pipe and making a pass if nothing is there. You need a lock on the write side of the pipe but not on the read side. I would not pass but a small struct across as you are still shared memory.

bluelasers:
FYI, I think we can go lockless on the high priority thread. We need a lock on the low priority side. (Issue #1678)

hzeller:
You transfer information between threads, you can't do this without locks. You need a memory barrier.

bluelasers:
You need to be able to control the meaning of the values and order of accesses. A lock is required serialize the memory accesses on the low side. A volatile variable can be used to control the memory accesses on the high side. A procedure will control meaning of the values. Fences/barrier and flushing may be required to make sure the values are in the correct order.

C/C++ did not provide this originally. GCC did in a semi-portable manner. Locks are the only way to really go about it in pthreads, if I understand it correctly. Pipe is expensive in some cases.

So I think I understand at least partly why it is what it is. Your saying it is still not a good idea?

hzeller:
Closing; as discussed it is a misguided attempt to solve a problem that does not exist.

PR #1660, opened 2024-05-29
Adds a python countdown program

vsumpi:
Fun little program that I wrote by my self. If not accepted any advice is welcome.


Ha2k4r (commented):
although this doesn't seem like something hzeller would add to the program it looks like you had a lot of fun making this and I like the idea keep making fun stuff!

bindings/python/countdown-n-lap.py
```diff
@@ -0,0 +1,81 @@
+'''
```
neuschaefer, line 1: Hello @vsumpi, if you look at the source tree, you'll see that there are already a bunch of examples written in Python, but they are in the `bindings/python/samples` directory, rather than in `bindings/python`

vsumpi: Okay, I have a question, how to set flags globally on my rpi so it is not embedded in the file, is it possible?

Issue #1663, opened 2024-06-11
Vmapper:Z requires physical re-mapping and rewiring of panels when switching from odd to even number in height


marcmerlin:
You can currently wire things so that they work whether you have odd or even, but if you try to add a panel, everything gets flipped around
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1014 still works, but is fairly suboptimal

marcmerlin:
![PXL_20240611_154150807](https://github.com/hzeller/rpi-rgb-led-matrix/assets/1369412/ec806716-aa68-49a6-ac04-26f5d2e2c88f)

![PXL_20240611_154215361](https://github.com/hzeller/rpi-rgb-led-matrix/assets/1369412/78074959-f22a-4ad5-ae50-18833251fc7a)

marcmerlin:
this is now fixed in https://github.com/hzeller/rpi-rgb-led-matrix/pull/1664

PR #1664, opened 2024-06-11
Fix for orientation being swapped as described in https://github.com/hzeller/rpi-rgb-led-matrix/issues/1663

marcmerlin:


marcmerlin:
After this fix

PXL_20240611_193703235

PXL_20240611_193738126

marcmerlin:
this is fixed, but I'll let @hzeller  push the close button since it's only fixed in my fork for now.

marcmerlin:
putting the issue in the subject line didn't work, let me reference it here: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1663

PR #1673, opened 2024-07-07
python: Don't depend on PIL/Pillow

neuschaefer:
As outlined in the commit message, the python binding does not really need to import PIL.


hzeller:
Thanks!

Issue #1671, opened 2024-06-30
Python bindings no longer build with 3.12


4censord:
When trying to build the python buildings with python3.12, one gets the following errors

<details>
<summary>The build error</summary

make -C ../../lib
make[1]: Entering directory 'rpi-rgb-led-matrix/lib'
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o gpio.o gpio.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o led-matrix.o led-matrix.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o options-initialize.o options-initialize.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o framebuffer.o framebuffer.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o thread.o thread.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o bdf-font.o bdf-font.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o graphics.o graphics.cc
g++ -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -fno-exceptions -std=c++11 -c -o led-matrix-c.o led-matrix-c.cc
cc  -I../include -W -Wall -Wextra -Wno-unused-parameter -O3 -g -fPIC  -DDEFAULT_HARDWARE='"regular"' -c -o hardware-mapping.o hardware-mapping.c
[...]

PR #1675, opened 2024-07-09
Regenerate cython files

neuschaefer:
Fixes: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1671


hzeller:
Seeing this, I think I'd be in favor if we switch to a mode where the generated cython files are not checked in, but part of the users' build.

Merging this now, but I think it would probably a better experience as stale generated code that caused the issue in the first place is not there to begin with.

Would you like to rework the Python binding for this ? (I suspect it mostly involves deleting the generated files and updating the Makefile and README.

(I don't use Python myself, so I would've never noticed)

Anyway, merging this for now, but I think the above suggestion might be a good next step.

hzeller:
I've put that in #1679

PR #1677, opened 2024-07-09
Fix Raspberry Pi model detection on mainline Linux

neuschaefer:
(please do not squash)

### DIFF
--- a/lib/gpio.cc
+++ b/lib/gpio.cc
@@ -244,6 +244,14 @@ static int ReadFileToBuffer(char *buffer, size_t size, const char *filename) {
   return r;
 }
 
+static int ReadBinaryFileToBuffer(uint8_t *buffer, size_t size, const char *filename) {
+  const int fd = open(filename, O_RDONLY);
+  if (fd < 0) return -1;
+  ssize_t r = read(fd, buffer, size);
+  close(fd);
+  return r;
+}
+
 /*
  * Try to read the revision from /proc/cpuinfo. In case of any errors, or if
  * /proc/cpuinfo simply contains zero as the revision, this function returns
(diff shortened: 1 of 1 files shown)


lib/gpio.cc
```diff
@@ -244,25 +244,54 @@ static int ReadFileToBuffer(char *buffer, size_t size, const char *filename) {
...
+  return pi_revision;
+}
+
+static uint32_t read_be32(const uint8_t *p) {
+  return p[0] << 24 | p[1] << 16 | p[2] << 8 | p[3];
+}
+
+static uint32_t ReadRevisionFromDeviceTree() {
+  char buffer[4];
+  if (ReadFileToBuffer(buffer, sizeof(buffer), "/proc/device-tree/system/linux,revision") < 0) {
```
hzeller, line 278: Note, `ReadFileToBuffer()` is meant to read a text file, so it reads one byte less than in the buffer and appends a nul-termination. So for the 4 byte linux revision to work , you need to give it a 5 byte buffer.

Also, you're checking for negative return value here, i.e. if it _wasn't_ able to read. Wouldn't you want to check for `== 4` here ?

hzeller:
Looks good in general, but I don't think it does what you think it does.

lib/gpio.cc
neuschaefer: Good points, I'll review the available I/O functions and test again

lib/gpio.cc
```diff
@@ -269,10 +269,26 @@ static uint32_t ReadRevisionFromProcCpuinfo() {
...
 
+static uint32_t read_be32(const uint8_t *p) {
+  return p[0] << 24 | p[1] << 16 | p[2] << 8 | p[3];
+}
+
+static uint32_t ReadRevisionFromDeviceTree() {
+  char buffer[4];
+  if (ReadFileToBuffer(buffer, sizeof(buffer), "/proc/device-tree/system/linux,revision") < 0) {
+    fprintf(stderr, "Failed to read revision from /proc/device-tree\n");
+    return read_be32((const uint8_t *)buffer);
```
neuschaefer, line 280: Another thing here: I've used a C-style cast, but I'm open to suggestions if you think there's a C++ cast that fits better

neuschaefer:
I think the logic is sound now, and I've checked that the revision is read correctly.

hzeller (approved):

PR #1679, opened 2024-07-10
Remove pre-generated cython files.

hzeller:
They easily get stale, and it is better to re-generate at compile time.


hzeller:
CC @neuschaefer to review.

neuschaefer:
Probably also a good idea:

```diff
diff --git a/shell.nix b/shell.nix
index 94d8a86..d19591e 100644
--- a/shell.nix
+++ b/shell.nix
@@ -11,5 +11,6 @@ pkgs.mkShell {
       ffmpeg
       pkg-config
       python3
+      python3Packages.cython
     ];
 }
```

hzeller:
> Probably also a good idea:  (add stuff to shell.nix)

Thanks. Also had to add distutils-extra; it looks like the distutils stuff is deprecated and removed in 3.12, but I don't know enough about Python to fix that. Up for refinement :)

Issue #1686, opened 2024-07-20
Panel P3 64x64 IC SM16208 SM5166  with matrix bonnet


LyubMaks:
Hello. I have a problem with 64x64 matrices with chips SM 6208 SM5166 scan 1/16 with strange addressing.
I made a video with an example of a pixel-mover (start x,y 0.0, then y - 0-63 and back, then x - 0-63).
Please help to make multiplexing for these panels

https://github.com/user-attachments/assets/e68b9cc2-b6b5-4e3f-b962-4b4186a380a0

https://github.com/user-attachments/assets/ac48bf31-194e-4d8d-a59d-0684579cd2f3

If you need to do additional tests, write what is needed.

LyubMaks:
![pixel_mover](https://github.com/user-attachments/assets/e2fb6859-1c7c-4771-a6e3-14c816ec9e65)
![ic](https://github.com/user-attachments/assets/4ae25628-324c-404b-9e0c-97faa063f276)

LyubMaks:
I was able to run this panel in another library (DMD_STM32) by changing some parameters (screen). But this library does not fit my needs. And I don't know how to use DMD_STM32 library settings for rpi-rgb-led-matrix library.
Maybe someone can integrate it.
![DMD_STM32](https://github.com/user-attachments/assets/9cd5d259-94df-4665-acf0-8a49699a38b5)

https://github.com/user-attachments/assets/31550e3b-d837-400c-9c8d-2abfe70563e7

TristanGrahn:
Ive also got a sm16208 p3 64x64 panel. And have been struggling to get it to display anything.

ixcue:
I have the same problem, please help. I already wrote to the author (@hzeller), but have not received a response yet...

davemaster:
Greetings, 

Have You MOD the active 3 board, to connect E-pin, or try --led-multiplexing = 0 .. to.. 18 (panel types)

davemaster:
Greetings,

Can You all share a close picture of your HUB75 connector? -INPUT-

GeorgeFlorian:
> Greetings,
> 
> Can You all share a close picture of your HUB75 connector? -INPUT-

I have made a separate issue: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1725

board707:
> I don't know how to use DMD_STM32 library settings for rpi-rgb-led-matrix library.
> Maybe someone can integrate it.

[...]

PR #1688, opened 2024-07-23
P3 RGB 64x64 1/16 scan Outdoor led panel multiplexer added

IGR2014:
Fix for issues #1631, #1686

By a scientific poking method the solution was found for this RGB LED matrix (see image)
Scan 1/16 (E line is GND-connected)
Used on RPi via Adafruit Bonnet

image

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -481,9 +481,9 @@ class P3Outdoor64x64MultiplexMapper : public MultiplexMapperBase {
 public:
   P3Outdoor64x64MultiplexMapper() : MultiplexMapperBase("P3Outdoor64x64MultiplexMapper", 2) {}
   // P3 RGB panel 64x64
-  // with pattern   [1] [3] [4] [7]
-  //                   |   /   |
-  //                [0] [2] [5] [6]
+  // with pattern   [1] [3]
+  //                 | \ |
+  //                [0] [2]
 
   void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
     const bool is_top_stripe = (y % (panel_rows_/2)) < panel_rows_/4;
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -487,11 +487,7 @@ public:
 
   void MapSinglePanel(int x, int y, int *matrix_x, int *matrix_y) const {
     const bool is_top_stripe = (y % (panel_rows_/2)) < panel_rows_/4;
-    const int quarter = x;
-    *matrix_x = ((2*quarter)
-                 + (is_top_stripe
-                    ? 1
-                    : 0));
+    *matrix_x = ((x*2) + (is_top_stripe ? 1 : 0));
     *matrix_y = ((y / (panel_rows_/2)) * (panel_rows_/4)
                  + y % (panel_rows_/4));
   }
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -477,6 +477,27 @@ public:
 };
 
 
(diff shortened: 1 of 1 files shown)


lib/multiplex-mappers.cc
```diff
@@ -477,6 +477,23 @@ class P10Outdoor32x16QuarterScanMapper : public MultiplexMapperBase {
 };
 
 
+class P3Outdoor64x64MultiplexMapper : public MultiplexMapperBase {
+public:
+  P3Outdoor64x64MultiplexMapper() : MultiplexMapperBase("P3Outdoor64x64MultiplexMapper", 2) {}
+  // P3 RGB panel 64x64
+  // with pattern   [1] [3] [4] [7]
```
IGR2014, line 484: Comment fix
```suggestion
  // with pattern   [1] [3]
  //                 | \ |
  //                [0] [2]
```

hzeller:
thanks, merged.

rotrico:
Hi IGR2014,
I bought same panel (I think that is same panel, I attach photo of pcb) from this store:
https://www.led-card.com/p3-outdoor-led-module.html

PCB chips have this text:
CHIPONE
ICND2153AP
061638MA
I tried your commit ( P3Outdoor64x64MultiplexMapper ) with this command:
./demo -D 0 --led-rows=64 --led-cols=64 --led-slowdown-gpio=4 --led-multiplexing=20
using a raspberry pi 4 running debian 11 with a board from electrodragon attached to row P0:
https://www.electrodragon.com/product/rgb-matrix-panel-drive-board-raspberry-pi/
PSU is a Meanwell LRS-350-5 tested with other setups.
But I can't see any pixel brighting, I tried --led-row-addr-type with values from 0 to 4 with same result
I discard problems with ribbon, psu, board...as if I connect other types of led panels (some P5 64x32) to same system they work flawless (obviously changing demo parameters).
In addition to your commit, maybe did you need to change anything in your hard in order to get your panel working? Or maybe I'm misssing something in parameters.
Also I contacted with seller asking for the panel datasheet, I'll attach if I get it.
Thanks.

IGR2014:
> PCB chips have this text: CHIPONE ICND2153AP 061638MA
Hello, maybe that's the problem. Possibly those chips require some other initialization code. Panels I used to work with were FM6126/FM6127 compatible

rotrico:
Ok, I put "ICND2153" in github search box and I found a project for arduino (https://github.com/CamelCaseName/HUB75Enano/tree/main) compatible with this chip.
I'll try to modify hzeller's library to add support for this chip looking chip datasheet and implementation in arduino library.
Also I got response from panel seller, they confirm that panel is ABCD and sent me hub75 pinout (nothing special despite confirming it's an ABCD, I attach here).
I'll tell if I can advance.
Thanks.
hub75 data sheet

rotrico:
I've spent last hours trying to understand the issue, If I'm right, IC's like ICDN2153 or ICN2053 have some sort of internal memory and PWM generators, a source: https://kuku.eu.org/?projects/hub75e/index , unlike other IC that only act as a buffers.
I'm not sure if this changes  (OE pin is replaced with GCLK pin and it has configuration registers) allow rpi-rgb-led-matrix to be compatible with those IC simply adding an initialization method like FM6126 / FM6127 or implies a completely rewrite of how to data is sent to panels.
I found an implementation for ESP32, but is only intended for one panel, I don't know how to write those configuration to all panels , also I tried to play with LE and GCLK signals in framebuffer.cc in a single panel to emulate esp32 init process with no result
There  is the  icn2053_start() method that shows how to init a panel using LE and GCLK signals:
https://bitbucket.org/nadyrshin_ryu/icn2053_esp32_demo/src/master/main/icn2053/icn2053.c
Tomorrow I'll try to find time to investigate more on this way.
Thanks.

PR #1709, opened 2024-09-24
Adding missing 'using' statements to the PlayGIF dotnet example

bacon1986:
I could build all the C# examples except the PlayGIF one because it was missing some 'using' statements so this fixes that. Thanks.


hzeller:
Thanks. I can't test it, but looks plausible.

PR #1727, opened 2024-11-04
.github: Update actions/checkout to v4

neuschaefer:
https://github.com/actions/upload-artifact warns:

> actions/upload-artifact@v3 is scheduled for deprecation on November
> 30, 2024. Learn more. Similarly, v1/v2 are scheduled for deprecation
> on June 30, 2024.


marcmerlin:
can you 
1) confirm this will not break users with older versions of git
2) explain what the new actions, add?
Basically I'd like to know what the gain is and the risk of breakage

neuschaefer:
1. This should not affect users at all, because it only applies to GitHub Actions, which are run on GitHub infrastructure. (It might theoretically affect some users which decide to run the Actions locally, but I've never heard such a thing)
2. Older versions are regularly deprecated in favor or newer versions; updates might be necessary in order to keep up with GitHub Actions' changing compatibility requirements. Other than that, the benefits are admittedly unclear. More information (albeit not much) is available at https://github.com/actions/checkout

marcmerlin:
Thanks for clearing things up, I should have noticed this only ran on github and not end users' repos.
From what I just saw now, this is only currently used to generate and test python bindings?

Are you working on those bindings? (if so, I can add you to the list of python knowledgeable people)
Were you upgrading the github actions proactively, or you noticed that the current bindings were not being built optimally and the new actions will do a better job?

neuschaefer:
> Thanks for clearing things up, I should have noticed this only ran on github and not end users' repos. From what I just saw now, this is only currently used to generate and test python bindings?

The purpose is essentially compile-testing: Making sure the code in pull requests doesn't have any obvious compile errors.

> Are you working on those bindings? (if so, I can add you to the list of python knowledgeable people) Were you upgrading the github actions proactively, or you noticed that the current bindings were not being built optimally and the new actions will do a better job?

I use the python bindings in a long-running project, but I'm not actively developing them.

I'm don't exactly remember what motivated me to make this PR, I think I just noticed the line in the file, or maybe there was a deprecation warning in the GitHub Actions user interface. Either way, the update was proactive, to push back the bit-rot.

PR #1737, opened 2024-12-08
Add method to create Font from (const) string

mcder017:
...rather than requiring fonts to be loaded through a File handle.  New ReadFont method is useful for allowing default built-in font simply using a header file with a defined string constant (e.g. BDF_10X20)

### DIFF
--- a/include/graphics.h
+++ b/include/graphics.h
@@ -21,6 +21,7 @@ struct Color {
   uint8_t r;
   uint8_t g;
   uint8_t b;
+  void setColor(uint8_t rr, uint8_t gg, uint8_t bb) {r = rr; g = gg; b = bb;}
 };
 
 // Font loading bdf files. If this ever becomes more types, just make virtual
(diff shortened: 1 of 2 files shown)


mcder017:
Users can then create a header file, such as bdf_10x20_local.h, that has content:

```
#ifndef BDF_10X20_LOCAL_H
#define BDF_10X20_LOCAL_H
const char* BDF_10X20_STRING = R"BDF_NON_KEYWORD(
---copy-paste the entire bdf font file here, from the STARTFONT line all the way through to the ENDFONT line---
)BDF_NON_KEYWORD";
#endif //BDF_10X20_LOCAL_H
```

and in the program itself, #include the header file and then load the font programmatically (no file access):
```
  rgb_matrix::Font smallFont;
  if (!myFont.ReadFont(BDF_10X20_STRING)) {
    fprintf(stderr, "Couldn't read internal font '%s'\n", BDF_10X20_STRING);  // should never happen if bdf font valid when pasted
  }
```

mcder017:
@hzeller merged in the recently approved 1-line change (to respect bbx offset) that WardBrian had resolved.

PR #1745, opened 2024-12-19
Removed unused function in Impl class

Tristan-Huen:
It seems that `RGBMatrix::Impl` has a declaration for a `Clear` function but no definition is found anywhere, and the function is never used.


hzeller:
Thanks, indeed that was a left-over from some much earlier implementation.

PR #1746, opened 2024-12-23
Added countdown with clock and sound

vsumpi:


marcmerlin:
@hzeller  this doesn't touch any code and adds an extra example, which is clearly safe.
Are you good with me merging that?

vsumpi:
> @hzeller this doesn't touch any code and adds an extra example, which is clearly safe. Are you good with me merging that?

Hi! I just want to make sure if I have to edit anything in the code. Like the setup phase or something? I'd be more than happy to assist. I'm a beginner and running a live project for biathlon and other events in Hungary based on this repo and codeblock. I'm using a Pi4+ 4gb and a Adafruit A2277 64x32 matrix with some speakers. :D

PR #1753, opened 2025-01-05
Python Packaging

chrisgilldc:
Per Marc's suggestion, making a PR here for review.
I put together packaging information for Python so that rpi-rgb-led-matrix can be installed as a package directly with pip. This is currently in a "works for me" state.
The build-rpi-rgb-led-matrix.py script uses the Hatch prebuild hooks to call the makefile and build the C libraries. The settings in hatch.toml make the Python package out of the files in "bindings/python/rgbmatrix". This does not include extensive checking for failures of the make process or passing of build options for different python packagers or cross-compiling.
Also note that PyPi no longer accepts local tags for version numbers. I had intended to put in code that would calculate a version based on the previous commit to the repo (ie: '0.0.1+20240924' for commit 0ff6a69). That may still be a viable strategy if you don't care about publishing to PyPi.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,13 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "rgbmatrix"
+version = "0.0.1+20240924"
+description = "Packaged version of Christoph Friedrich's rpi-rgb-led-matrix library."
+authors = [{name="Christoph Friedrich", email="christoph.friedrich@vonaffenfels.de"}]
+maintainers = [{name="Chris Gill", email="chris@chrisgill.net"}]
+license = "GPL-2.0-or-later"
+readme = "README.md"
+requires-python = ">=3.11"
--- a/README.md
+++ b/README.md
@@ -1,3 +1,14 @@
+RGB LED Python Package
+======================
+This fork was created to smooth installation of the Python rpi-rgb-led-matrix
+bindings (rgbmatrix). Wheel is built on Python 3.11 on a Pi3 - which should work
+for all intended use cases but has not been thoroughly tested. I'm purely an
+amateur so welcome someone with more skill and experience picking this up and
+improving it.
+
+As the originating library does not actually version, the package version is using
+the package's version (0.0.1) plus the date of the last commit as the version.
+
 Controlling RGB LED display with Raspberry Pi GPIO
 ==================================================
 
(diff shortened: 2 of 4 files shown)


hzeller:
I personally don't use Python, so have no idea what an expected packaging in the Python-world would be.
Maybe @Saij has an opinion and can review this PR ?

marcmerlin:
@ty-porter would you be able to weigh in on this? 
Since we have no project admin currently using python, I'm happy if you two guys help out and work on adding python improvements. 
If you guys both agree on changes, I'm happy to rubber stamp them and commit them

pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "rgbmatrix"
+version = "0.0.1+20250323"
```
ty-porter: I think we would want to avoid hard-coding this, since it would require an update to the package version each time a new commit is merged. Forgetting to update this value is really bad, since it means pip might not pick up changes.

We should look at [hatch-vcs](https://github.com/ofek/hatch-vcs) to version based on git SHA instead. I haven't used this personally, but sounds like it does exactly what we need:

```
  [build-system]
  requires = ["hatchling", "hatch-vcs"]

  [tool.hatch.version]
  source = "vcs"
```

This should let us dynamically define the version, something like `0.0.1+g8f3a1c2`. We'll need to check it -- I believe it uses latest tag + SHA by default.

marcmerlin:
Please let me know if you guys agree, and whether this should be merged or not.
@ledvinap  please have a look too.

marcmerlin:
@chrisgilldc  @ty-porter  can you check where https://github.com/hzeller/rpi-rgb-led-matrix/pull/1477 fits in all this?

marcmerlin:
I closed this suggestion as per @ledvinap  not sure if anything in it was useful
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1457

marcmerlin:
Apologies, I got the PR and the bug confused, will move my comments to the bug since hopefully this PR will get reviewed by you guys, approved and submitted.

marcmerlin:
To be clear, I'm ok with merging this but cannot review it, so I would need a pretty strong thumbs of from both of you that it's ok to merge as is. It does not need to be perfect or final, just better than the current situation.

ty-porter:
This looks good enough for now. We can tweak versioning later as needed. @chrisgilldc might want to update the version one last time before merging.

marcmerlin:
thank you to both. I'm definitely a "better is the enemy of good", so as long as it's an improvement and it works, let's get it in, which is now done.

PR #1759, opened 2025-02-06
Fixes simple square not staying on when running demo

Heher:
The third demo, simple square, exist immediately when you run it because it doesn't wait for the interrupt. This adds it in so it shows until you stop it.


marcmerlin:
confirmed new code works as intended, thank you

PR #1761, opened 2025-02-25
Clarify README to note new boot config & cmdline file locations on bookworm

sciguy14:
On the bookworm release of Raspbian the config.txt and cmdline.txt files were moved from `/boot` to `/boot/firmware`. They are not symlinks for backwards compatibility, so I've updated the README to make it clear that the new file locations are the ones that need to be modified if you are on a newer Raspbian build.

See here: https://github.com/RPi-Distro/raspberrypi-sys-mods/issues/88


marcmerlin (approved):
thank you

PR #1762, opened 2025-02-27
Fix MemStreamIO memory leak, compatability for CMake and add ability to mock RPI hardware

sshcrack:
This pull request makes several important changes to the `rpi-rgb-led-matrix` project, including setting up the CMake build system, fixing a memory leak caused by class MemStreamIO and introducing conditional compilation for mocking Raspberry Pi hardware. Below are the key changes grouped by theme:

### Build System Setup:
* Added a `CMakeLists.txt` file to set up the CMake build system, specifying the minimum required version, project name, C++ standard, and handling source files and include directories. It also includes conditional compilation for mocking Raspberry Pi hardware.

### Class Enhancements:
* Added a `Clear` method and a destructor to the `MemStreamIO` class in `include/content-streamer.h` and implemented these methods in `lib/content-streamer.cc`.

### Conditional Compilation for Mocking:
* Introduced `#ifndef MOCK_RPI` and `#ifdef MOCK_RPI` directives in various methods in `lib/content-streamer.cc` and `lib/led-matrix.cc` to allow for conditional compilation when mocking Raspberry Pi hardware.

### Fixed a memory leak
* Replaced the `CreateLuminanceCIE1931LookupTable` function with a `ColorLookupManager` singleton class to manage the luminance lookup table in `lib/framebuffer.cc`.

### Code Cleanup:
* Added a `#pragma once` directive to the `include/content-streamer.h` file to prevent multiple inclusions.


CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.5.0)
+project(rpi-rgb-led-matrix VERSION 0.1.0)
+
+set(CMAKE_CXX_STANDARD 14)
+
+file(GLOB_RECURSE SHARED_SOURCES
+        lib/*.c*
+)
```
sshcrack, line 8: This can also list every source file, but I thought it would be easier for maintanance just to leave it as a glob

examples-api-use/image-example.cc
```diff
@@ -82,14 +82,18 @@ static ImageVector LoadImageAndScaleImage(const char *filename,
...
   // Copy all the pixels to the canvas.
+  // Get direct access to pixel data
+  const Magick::PixelPacket *pixels = image.getConstPixels(0, 0, image.columns(), image.rows());
+
   for (size_t y = 0; y < image.rows(); ++y) {
+    const Magick::PixelPacket *row = pixels + (y * image.columns());
     for (size_t x = 0; x < image.columns(); ++x) {
-      const Magick::Color &c = image.pixelColor(x, y);
-      if (c.alphaQuantum() < 256) {
+      const auto &[blue, green, red, opacity] = row[x];
```
hzeller, line 91: This http://www.graphicsmagick.org/Magick++/PixelPacket.html

describes the PixelPacket to contain RGBA ... by you seem to deconstruct it as BGRA

In fact, it can be either https://foss.heptapod.net/graphicsmagick/graphicsmagick/-/blob/branch/default/magick/image.h?ref_type=heads#L529-548

So I don't think it is safe to do it this way, but rather access the struct directly.

Also, is the raw sample the raw row contains the same as a Quantum (as we use that assumption to scale).

hzeller (commented):
Not sure what leak you saw in `MemStreamIO`, but the std::string it has does cleanup already.

I am not convinced yet that the `MOCK_RPI` is good when it requires to have attention where to put the `#ifdef`s, so I think that should be broken out into a separate change to discuss. Rough idea would be to have a different `Impl` which then could be `#ifdef`-swapped.

I am wondering though what your intention is, as you also had some changes in the content streamer with regard to the `MOCK_RPI`, but that should typically not be needed (for instance, you can just use the image-viewer to create a an image stream and run it on your PC where you don't need access to the Raspberry Pi (the `-O` option in the led-image-viewer).

I think the cmake change should be a separate pull request.

Also the `ColorLookupManager` as a separate pull request. I'd call it `ColorLookupTable` as that is what it is ('Manager' in any class name always means that 'naming-is-hard' hit again :) ).

So
  * Make ColorLookupTable a separate pull request; that change looks good
  * Cmake ... yeah I could be convinced, but then we should also make everything compatible with CMake, i.e. build the Python bindings with it.
  * `MOCK_RPI` requires a separate pull request and possibly some discussion how to best do that.
  * The PixelPacket change looks like a breaking change to me or 'accidentally' working sometimes. Not sure if it is worth.

sshcrack:
I‘m splitting the PR into the smaller chunks
- [Add cmake build system](https://github.com/hzeller/rpi-rgb-led-matrix/pull/1764)
- [Singleton for Color Lookup](https://github.com/hzeller/rpi-rgb-led-matrix/pull/1763)
- Mock RPI
- Maybe MemStreamIO fix (not at home currently, will review changes)

sshcrack:
About the MemStreamIO: I had heaptrack running for about 6 hours and saw leaks coming from the string deallocation. Maybe I had messed up implementing some functionality which caused memory leaks. Adding a destructor with a specific clear instruction seemed to fix the deallocation issues. Maybe wrong tests on my end though

sshcrack:
Closing this PR as I‘ve split this one up

PR #1763, opened 2025-02-28
Fix memory leak for color lookup

sshcrack:
This PR adds a singleton that is used to get a color lookup

### DIFF
--- a/lib/framebuffer.cc
+++ b/lib/framebuffer.cc
@@ -599,19 +599,28 @@ static uint16_t luminance_cie1931(uint8_t c, uint8_t brightness) {
 struct ColorLookup {
   uint16_t color[256];
 };
-static ColorLookup *CreateLuminanceCIE1931LookupTable() {
-  ColorLookup *for_brightness = new ColorLookup[100];
-  for (int c = 0; c < 256; ++c)
-    for (int b = 0; b < 100; ++b)
-      for_brightness[b].color[c] = luminance_cie1931(c, b + 1);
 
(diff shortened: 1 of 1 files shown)


lib/framebuffer.cc
```diff
@@ -589,29 +589,37 @@ void Framebuffer::Clear() {
...
+      static ColorLookupTable instance;
+      return instance.lookups_[brightness - 1];
+    }
 
-static inline uint16_t CIEMapColor(uint8_t brightness, uint8_t c) {
-  static ColorLookup *luminance_lookup = CreateLuminanceCIE1931LookupTable();
-  return luminance_lookup[brightness - 1].color[c];
-}
+    // Do CIE1931 luminance correction and scale to output bitplanes
+    static uint16_t luminance_cie1931(uint8_t c, uint8_t brightness) {
```
hzeller, line 604: This looks like it wants to be a private member as well. The public interface only needs `GetLookup()`

sshcrack: Alright, didn't know the usage outside of the library so I made it public, will move to private for now

marcmerlin:
@hzeller  you're ok with merging this?

marcmerlin:
I would like to review this but not super obvious in just a couple of minutes when I'm going through dozens or stalled PRs, would still like @hzeller  to review this

hzeller (approved):

hzeller:
Thanks for the ping. Merged.

PR #1765, opened 2025-02-28
Add ability to mock RPI hardware

sshcrack:
This will add the ability to mock RPI hardware. Mocking RPI hardware can be especially useful when the project is large and needs a lot of compile time, especially on the slow RPI 3 hardware. Building and debugging the program on a computer can speed up development significantly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,6 @@ set(CMAKE_CXX_STANDARD 14)
 file(GLOB_RECURSE SHARED_SOURCES
         lib/*.c*
 )
-if(MOCK_RPI)
-    add_definitions(-DDISABLE_HARDWARE_PULSES)
-    add_definitions(-DMOCK_RPI)
-endif()
 
 add_library(${PROJECT_NAME} SHARED)
 target_sources(${PROJECT_NAME}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,11 @@ file(GLOB_RECURSE SHARED_SOURCES
         lib/*.c*
 )
 
+if(MOCK_RPI)
+    add_definitions(-DDISABLE_HARDWARE_PULSES)
+    add_definitions(-DMOCK_RPI)
+endif()
+
 add_library(${PROJECT_NAME} SHARED)
 target_sources(${PROJECT_NAME}
         PRIVATE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,31 @@
+cmake_minimum_required(VERSION 3.5.0)
+project(rpi-rgb-led-matrix VERSION 0.1.0)
+
+set(CMAKE_CXX_STANDARD 14)
+
+file(GLOB_RECURSE SHARED_SOURCES
+        lib/*.c*
+)
+if(MOCK_RPI)
+    add_definitions(-DDISABLE_HARDWARE_PULSES)
+    add_definitions(-DMOCK_RPI)
+endif()
+
+add_library(${PROJECT_NAME} SHARED)
+target_sources(${PROJECT_NAME}
+        PRIVATE
+        ${SHARED_SOURCES}
+)
+# Main Target Compile
+
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_14)
+target_include_directories(${PROJECT_NAME}
+        PRIVATE
+        # where the library itself will look for its internal headers
+        ${CMAKE_CURRENT_SOURCE_DIR}/lib
+        PUBLIC
+        # where top-level project will look for the library's public headers
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        # where external projects will look for the library's public headers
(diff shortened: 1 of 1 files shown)


sshcrack:
Also I‘m quite unsure how to implement this. I think creating a new implementation is a good idea but should we make the RgbMatrix::Impl virtual so a mocking class can overwrite these methods?

sshcrack:
Also are there any emulators for seeing output of the led matrix? Like in the python implementation?

hzeller:
I usually program all outputs against a Canvas implementation. Having said that, doing full animations with wait for sync and stuff is not possible with that.

We could consider virtual methods, but given that this might really be a compile-time choice (testing on an animation vs. 'production'), this might be unnecessary overkill. Let me think about this more.

Do you have a link to a project where you'd like to use this ?

sshcrack:
I'm using this for my led matrix application which has even support for plugins: [Repo](https://github.com/sshcrack/led-matrix)

Tristan-Huen:
> Also are there any emulators for seeing output of the led matrix? Like in the python implementation?

I managed to make my own version using @hzeller's flaschen-taschen, ripping the code from the terminal plotter, and modifying the underlying matrix code. Currently it uses a run-time flag like the other options but would be way better as a compile-time choice imo. It's not the best and is really just a hacked-together thing I did out of need (funnily enough my project that I am using it for is the exact same concept as yours).

WindowsTerminal_bvHI0b2CqS

sshcrack:
Would be great if you could upload the code. I don't know if I'll be able to fix the missing features but I will try.

sshcrack:
Okay, I've created a simple emulator, again there is no support for pixel mapping but  it worked for my use case!

sshcrack:
Alright it works on my machine™

CMakeLists.txt
sshcrack, line 1: Added this file for now, don't know about the other pull request for CMake

Tristan-Huen:
Oh damn it seems you already implemented it from scratch by the time I saw this. Just from initial skimming it seems like you went with SDL which is actually nice since it removes the annoyances with the terminal (e.g. scaling). 

> Would be great if you could upload the code. I don't know if I'll be able to fix the missing features but I will try.
> 
> Alright I looked at the source code and I don't really understand how pixels are stored in bitplanes etc. It would be great if you could take some time!

The bitplane logic is something I don't understand fully as well. For the pixel mapping, another simulator I made using SFML gets around this by creating the window in the `Impl::UpdateThread::Run` function since the changes to the `shared_pixel_mapper_` variable of the `Impl` class gets updated in the constructor by the calls to `ApplyPixelMapper` and `ApplyNamedPixelMappers` at the end:

https://github.com/hzeller/rpi-rgb-led-matrix/blob/d10f518f291bd34ca4816ef25cadf2446b882662/lib/led-matrix.cc#L369-L403

sshcrack:
Yeah, I just ended up implementing a virtual frame buffer which the emulator update thread can read easily

sshcrack:
Oh, didn't mean to push this to this PR, will undo

WardBrian:
> Also are there any emulators for seeing output of the led matrix? Like in the python implementation?

https://github.com/ty-porter/RGBMatrixEmulator

sshcrack:
> > Also are there any emulators for seeing output of the led matrix? Like in the python implementation?
> 
> https://github.com/ty-porter/RGBMatrixEmulator

I should've clarified: I meant if there is a cpp based emulator so outputs of a cpp program can be seen

marcmerlin:
This looks interesting, but indeed I'm going to leave it to @hzeller 
For the C interface, I handled this a different way. I have a framebuffer::gfx 2D API which I recommend since it works against many backends, and I added a linux/sdl backend to display directly on my laptop without hardware:
https://marc.merlins.org/perso/arduino/post_2020-01-24_Running-Arduino-code-with-2D-FastLED_-Adafruit_GFX_-and-LEDMatrix-displays-on-Linux.html

then the same code can run on rpi with this lib:
https://marc.merlins.org/perso/arduino/post_2020-01-01_Running-FastLED_-Adafruit_GFX_-and-LEDMatrix-code-on-High-Resolution-RGBPanels-with-a-Raspberry-Pi.html

marcmerlin:
@hzeller are you ok giving a +1 to this?

marcmerlin:
I just did a big push to merge or close all old PRs.
this one looks interesting, clearly a lot of work went into it, but it's also big to merge and I'm not sure about the current need, especially since I mentioned in 
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1765#issuecomment-2840837520
that I have my own way to mock hardware (basically I have a framebuffer class that can display on linux) and is a lot less work/impact to the lib (basically 0) as well as offers displaying on neomatrix arrays, TFTs, and other chips with 0 work for the user.

It does not look like @hzeller  will have time to reasonably review this or get involved, and neither will I.
I don't want to shut down good work, but also want to be realistic.
If you can update this and commit to keeping it updated (because honestly I won't be able to), unless it should just keep working with close to 0 effort, I'm willing to consider it, but if not, for now I'm going to close it.
You can update this PR and update it, and it can be re-opened/considered, I just want to keep the immediate PR todo list as small as possible.

PR #1776, opened 2025-03-24
Add ability to query font character width to led-matrix-c

### LINKED ISSUE #1775: Provide CharacterWidth font function in the C api
```c
// file: include/led-matrix-c.h
int character_width_font(struct LedFont *font, uint32_t unicode_codepoint);

// file: lib/led-matrix-c.cc
int character_width_font(struct LedFont *font, uint32_t unicode_codepoint) {
  return to_font(font)->CharacterWidth(unicode_codepoint);
}
```

I think someone can use something like utf8.h to do codepoint separation, so there's no need to expose the internal `utf8_next_codepoint` function for this to be useful.

WardBrian:
This would close #1775

### DIFF
--- a/include/led-matrix-c.h
+++ b/include/led-matrix-c.h
@@ -391,6 +391,9 @@ int baseline_font(struct LedFont *font);
 // Read the height of a font
 int height_font(struct LedFont *font);
 
+// Get the width of a specific codepoint in the given font
+int character_width_font(struct LedFont *font, uint32_t unicode_codepoint);
+
 // Creates an outline font based on an existing font instance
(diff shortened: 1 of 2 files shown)


Ministorm3:
This wouls be a super useful addition. I currently manually parse the character width with my own implementation. Since the library already handles height, I see no reason why it shouldn't also do width.

@hzeller any chance we can get this reviewed?

marcmerlin (approved):
I'll admit that I don't know much about fonts, but it's also obvious that this can't break anything, so approved :)

PR #1777, opened 2025-03-25
Fix a number of compiler warnings and CodeQL findings

IngmarStein:
```
clock.cc: In function ‘int main(int, char**)’:
clock.cc:196:14: warning: ignoring return value of ‘ssize_t write(int, const void*, size_t)’, declared with attribute warn_unused_result [-Wunused-result]
  196 |   (void)write(STDOUT_FILENO, "\n", 1);  // Create a fresh new line after ^C on screen
      |         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~

rpi-rgb-led-matrix/lib/framebuffer.cc:299:
Multiplication result may overflow 'int' before it is converted to 'unsigned long'.
This rule finds code that converts the result of an integer multiplication to a larger type. Since the conversion applies after the multiplication, arithmetic overflow may still occur.

rpi-rgb-led-matrix/lib/framebuffer.cc:845:
Comparison between  of type uint8_t and  of wider type const int.
In a loop condition, comparison of a value of a narrow type with a value of a wide type may result in unexpected behavior if the wider value is sufficiently large (or small). This is because the narrower value may overflow. This can lead to an infinite loop.

examples-api-use/demo-main.cc:270
Multiplication result may overflow 'int' before it is converted to 'size_t'.
This rule finds code that converts the result of an integer multiplication to a larger type. Since the conversion applies after the multiplication, arithmetic overflow may still occur.

examples-api-use/ledcat.cc:43
Multiplication result may overflow 'int' before it is converted to 'ssize_t'.
This rule finds code that converts the result of an integer multiplication to a larger type. Since the conversion applies after the multiplication, arithmetic overflow may still occur.
```

### DIFF
--- a/examples-api-use/clock.cc
+++ b/examples-api-use/clock.cc
@@ -14,7 +14,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-#include <unistd.h>
+#include <iostream>
 
 #include <vector>
 #include <string>
@@ -193,6 +193,6 @@ int main(int argc, char *argv[]) {
   // Finished. Shut down the RGB matrix.
   delete matrix;
 
-  write(STDOUT_FILENO, "\n", 1);  // Create a fresh new line after ^C on screen
+  std::cout << std::endl;  // Create a fresh new line after ^C on screen
   return 0;
 }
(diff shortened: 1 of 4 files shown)


marcmerlin:
thanks, I tested this against an old raspbian 4 and new dietpi 6 (32 and 64bit) and both built fine, so that looks good to me.

PR #1793, opened 2025-06-10
Fix potential integer overflow in Timers::sleep_nanos

jonmorrissmith:
The issues are:

1. The subtraction (after - before) is cast to uint32_t which can wrap around if after is less than before (due to the timer wrapping around)

2. Multiplying by 1000 could overflow a 32-bit integer for large time differences

3. The comparison nanoseconds_passed > nanos might not work correctly with wrapped values

Key Improvements

1. Handling Timer Wraparound: The code now explicitly checks if after is less than before, which indicates the timer has wrapped around. In that case, it calculates the elapsed time correctly.

2. Avoiding Arithmetic Overflow: I've used uint64_t for the calculation of nanoseconds_passed to ensure that the multiplication by 1000 doesn't overflow.

3. Safe Comparisons: The comparison with nanos now uses proper type casting to avoid potential issues with signed vs. unsigned comparisons.

4. Underflow Protection: Added a safety check to ensure we don't pass a negative value to the busy wait function if any calculation error occurs.

This should handle timer wraparound correctly and avoid potential integer overflows, making the timing system more robust, especially in cases where displays might run for extended periods.

### DIFF
--- a/lib/gpio.cc
+++ b/lib/gpio.cc
@@ -545,42 +545,58 @@ static uint32_t JitterAllowanceMicroseconds() {
 }
 
 void Timers::sleep_nanos(long nanos) {
-  // For smaller durations, we go straight to busy wait.
+    // For smaller durations, we go straight to busy wait.
+    // For larger duration, we use nanosleep() to give the operating system
+    // a chance to do something else.
 
-  // For larger duration, we use nanosleep() to give the operating system
-  // a chance to do something else.
-
-  // However, these timings have a lot of jitter, so if we have the 1Mhz timer
-  // available, we use that to accurately mesure time spent and do the
-  // remaining time with busy wait. If we don't have the timer available
-  // (not running as root), we just use nanosleep() for larger values.
-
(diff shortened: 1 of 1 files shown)


hzeller:
`before` and `after` are unsigned, so they will behave _exactly_ as we want on wrap-around: the difference will be the time we're looking for. If before was before wrapping around and after after wrapping around, the difference is exactly the number of microseconds passed. No need to manually special-case that.

Before 1000x the microseconds we observe creates an issue, more than 4 million microseconds have passed,, that is 4 seconds. At most, we use this function to sleep for 1 millisecond.

So you're trying to solve an issue that is not there, but making the code harder to understand in the process.

jonmorrissmith:
Thanks for the detailed explanation - my misunderstanding on the context.

hzeller:
In any case: thanks for looking out for potential problems. Even if in this case it was not a big problem to solve, having a sharp eye reviewing and observing potential issues in code is an important work and skill.

PR #1814, opened 2025-10-17
Updating Python binding setup.py imports to use setuptools, enabling use of Python ≥ 3.12

gidger:
Distutils was removed in Python 3.12, setuptools provides the needed functionality.

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,9 +1,7 @@
 #!/usr/bin/env python
 
-# The following distutils seems to stop working with newer
-# Python version >= 3.12
-# Someone with Python knowlege, please fix.
-from distutils.core import setup, Extension
+# 2025-10, updated to import from setuptools rather than distutils.core which was removed in Python 3.12.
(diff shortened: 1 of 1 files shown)


marcmerlin:
Can you confirm that this does not break with older python versions as long as they are not way too old?  I want to make sure I'm not breaking existing users by merging this.

gidger:
Shouldn't be any issues for anything but **extremely** old versions of Python. Distutils phase out began in 2014.

See the [relevant PEP](https://peps.python.org/pep-0632/#migration-advice) for more details. And additional details on setuptools [here](https://setuptools.pypa.io/en/latest/setuptools.html).

Distutils as a standalone was considered depreciated as of Python 3.10 and was removed entirely in 3.12. Setuptools is an enhanced alternative to distutils that includes a complete copy of distutils. The suggested migration path is to use the equivalent imports from setuptools, replacing those from distutils.

marcmerlin:
thanks for confirming

PR #1818, opened 2025-10-27
Refactor broken unsafe SetImagePillow()

### LINKED ISSUE #1769: Python bindings: Matrix.SetImage uses deprecated Pillow APIs
The default call (passing `unsafe=True`) leads to the warning `unsafe_ptrs property is deprecated and will be removed in Pillow 12`

More information in the [Pillow documentation](https://pillow.readthedocs.io/en/stable/releasenotes/11.0.0.html#get-internal-pointers-to-objects), which suggests `Image.Image.getim()` for interacting with C code

ty-porter:
Supersedes #1817 and resolves #1769

Previous PR was broken -> `image.im.getim()` should have been `image.getim()`

Second, using PyCapsules means we need precise byte offsets into the image structs (in this case `ImagingMemoryInstance` they contain in order to access the image frame. `get_unsafe_ptrs()` did this for us previously. 

https://github.com/python-pillow/Pillow/blob/e36e67081ae9d4162f5254ce6f3c1478fc342d9c/src/libImaging/Imaging.h#L80

Unfortunately it's not trivial, the first attribute `mode` specifically is not stable and is changing from `char[]` to `struct*` to enum value between Pillow 11 / 12, so the offsets are not stable.

Cython also cannot compile the header we need natively:

```python
# doesn't work because Cython can't handle typedef *
cdef extern from "Imaging.h":
    cdef struct ImagingMemoryInstance:
        pass # Filled out from header
```

Instead, this PR creates a simple C shim that hides `ImagingMemoryInstance` and provides portable access to the `image32` field needed for fast `SetImage()`

Retested this change on a fresh install of Pi OS Bookworm, I think the fact the previous PR worked was a competing installation.

### DIFF
--- a/bindings/python/rgbmatrix/core.pyx
+++ b/bindings/python/rgbmatrix/core.pyx
@@ -14,7 +14,6 @@ cdef extern from "shims/pillow.h":
 @cython.boundscheck(False)
 @cython.wraparound(False)
 cdef int** get_pillow_buffer(object capsule):
-    cdef void *raw_ptr
     cdef void *image
 
     image = PyCapsule_GetPointer(capsule, PyCapsule_GetName(capsule))


MikeRogers27:
Which version of pillow should I be installing?  <12 or is 12 and above fine with this PR?

ty-porter:
<12 is compatible prior to this PR. After this, all versions should be

MorganKryze:
Hi! I am facing this issue and am curious about when you think the PR could be merged?
Thanks for your time!

WeegeeNumbuh1:
Also awaiting this PR merge. Recently built a new RGB matrix system using the latest Raspberry Pi OS and found out a part of one of my projects no longer works due to PIL 12.0.

This fix will also benefit the people that use the Adafruit guide for implementing this library with Python as PIL is now at 12 and the older commits that their build script pulls will no longer work moving forward.

marcmerlin:
Just  to put things in context, I didn't write this driver, and I do not use its python interface, at all.  I wanted @hzeller  to have a chance to look at this PR, but my understanding is that he doesn't really have time to work on this code, so I'm going to trust your work and merge this.

marcmerlin:
Everyone affected, please confirm head now fixes your issues with this.

WeegeeNumbuh1:
> Everyone affected, please confirm head now fixes your issues with this.

@marcmerlin I just built it and can confirm this merge fixes the issue and it works on my end (to the extent that my project uses these bindings)

marcmerlin:
Perfect, thank you both

hzeller:
Thanks for merging @marcmerlin !
The Python code is an external contribution and I also have never used it, nor can I read Python code; so no need to wait for me for merging.

marcmerlin:
@ty-porter since no one with commit access is actually  using the python bindings, would you be interested in being the python binding maintainer for this project?   We can also go in the commit logs and see how else submitted the original bindings and recent fixes.

ty-porter:
@marcmerlin Sure, I'm happy to help out.

marcmerlin:
Thanks. Let's start with 
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1753

PR #1819, opened 2025-10-30
New DoubleZ multiplexer. Used for 64x32 HUB75 RGB with ICN2037BP chip.

xc218m:
IMG_20251030_120456
IMG_20251030_120506
IMG_20251030_120511

### DIFF
--- a/lib/multiplex-mappers.cc
+++ b/lib/multiplex-mappers.cc
@@ -493,6 +493,36 @@ public:
   }
 };
 
+class DoubleZMultiplexMapper : public MultiplexMapperBase {
+  public:
+    DoubleZMultiplexMapper() : MultiplexMapperBase("DoubleZ", 2) {}
+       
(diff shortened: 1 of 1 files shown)


marcmerlin (approved):

marcmerlin:
I can't test or review this, but it's self contained, so I trust it works and I know it won't break anyone else  :)

xc218m:
Yes, it works. Tested with a single panel and 2x2 configuration on raspberry pi 4 with these options:
--led-cols=64 --led-rows=32 --led-row-addr-type=0 --led-multiplexing=21 --led-slowdown-gpio=5 --led-no-drop-privs --led-chain=4 --led-parallel=1 --led-pixel-mapper="U-mapper"
Originally the display was used with Huidu E62 controller. It worked fine with 1x4, but 2x2 caused second line to have some shifted pixels.

hzeller:
Thanks!

PR #1827, opened 2025-11-27
Re-use CreateFrameCanvas() in samples/gif-viewer.py

redphx:
> Previously https://github.com/hzeller/rpi-rgb-led-matrix/pull/1556

I used the original sample to display a list of GIF files.  

After a few minutes I got this warning:  
```
CreateFrameCanvas() called 500 times. Usually you only want to call it once (or at most a few times) for double-buffering...
```

I modified the sample and re-use the canvas instead.


marcmerlin:
thanks again for the fix.

PR #1832, opened 2025-12-04
[Bindings/C#] Upgrade to .NET 8.0

smdn:
The codes under `bindings/c#/` uses .NET 6.0.
However, .NET 6.0 has already reached the end of its lifecycle, requiring migration to either .NET 8.0 or 10.0.

This PR changes the `TargetFramework` to `net8.0` to use .NET 8.0.

.NET 10.0 has already been released, but the apt package `dotnet10` is not yet available, requiring a more complex installation instruction.
Therefore, this PR changes to use .NET 8.0, which can be easily installed using the `apt install` command.

### DIFF
--- a/bindings/c#/README.md
+++ b/bindings/c#/README.md
@@ -8,7 +8,7 @@ To build the C# wrapper for the RGB Matrix C library you need to first have __.N
 
 ### Install .NET SDK
 
-`sudo apt install dotnet6` should work in most cases.  
+`sudo apt install dotnet8` should work in most cases.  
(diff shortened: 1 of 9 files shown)


gingters:
Probably not the right place, but we should think about what .NET versions should be supported for how long in general, and what .NET version should be used for development.
Usually, the new .NET Versions are generally backwards-compatible, so something compiled for .NET 8 should also work flawlessly on 9 and 10. We can also target backwards (using .NET 10 to build packages compatible with .NET 8). Newer versions also are usually faster than the previous ones, and usually also introduce more memory / allocation optimizations, which should benefit the usage on the Pi. So the newer, the better.

Issue #1749, opened 2025-01-01
Python Packaging


chrisgilldc:
Good afternoon!

In the effort to package some of my own python projects, it was frustrating me I couldn't include 'rgbmatrix' as a dependency.

I've worked up a fork with a pyproject.toml definition, build script, etc. I've tested it on my development Pi3 and it all seems to work fine.

Due to versioning limitations it can't be published to PyPi, but I've tested installing with pip straight from git and it works. Fork is [here](https://github.com/chrisgilldc/rpi-rgb-led-matrix) and hope it's helpful.

marcmerlin:
Hi @hzeller, I know nothing of the python bindings, what do you feel is the best course of action?
Something to potentially merge, reference, or other?
@chrisgilldc I can't guarantee it'll be merged, but can you send a PR from your tree to this project so that Henner has a chance to review and potentially, merge?

chrisgilldc:
@marcmerlin Just noting here I did create that PR back on the 5th, it's #1753

marcmerlin:
Adding @py-porter and @ledvinap  from https://github.com/hzeller/rpi-rgb-led-matrix/pull/1753
as mentioned there
I'm happy to have the 3 of you be in charge of python updates, and if 2 out of 3 agree on a change, to merge it (3 out of 3 is better, but maybe only 2 of you are available)

I will add other bugs/PRs to review here

marcmerlin:
I closed this suggestion as per @ledvinap not sure if anything in it was useful
#1457

marcmerlin:
@chrisgilldc @ty-porter can you check where #1477 fits in all this?

marcmerlin:
Also please have a look at https://github.com/hzeller/rpi-rgb-led-matrix/pull/1257/files from @mostthingsweb and recommend whether we should look/consider or if I can close.

marcmerlin:
@ty-porter  and @ledvinap  can you add yourselves as assignees of this issue so you get updates and can help/chime in on python updates?

marcmerlin:
Another PR to look at if you find it helpful to integrate:
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1257

marcmerlin:
I'll merge simple standalone python examples
https://github.com/hzeller/rpi-rgb-led-matrix/pull/909

chrisgilldc:
[...]

PR #1833, opened 2025-12-09
Revert distutils re-addition over setuptools

ty-porter:
Addresses new issue found in #1749

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -1,9 +1,6 @@
 #!/usr/bin/env python
 
-# The following distutils seems to stop working with newer
-# Python version >= 3.12
-# Someone with Python knowledge, please fix.
-from distutils.core import setup, Extension
+from setuptools import setup, Extension
 
 core_ext = Extension(
     name                = 'core',


rwstca:
Just to confirm, still works for Python < `3.11`?

Thanks from another Python bindings user!

ty-porter:
Yes, `setuptools` is supported for all current Python versions (at least 3.10+)

rwstca:
I don't want to overwhelm the scope of this small change, but I had an idea I wanted to throw out there for discussion. This `setup.py` file I have been lead back to as a source of bugs a number of times. I wonder since we are using the modernized `pyproject.toml` in the root project folder now whether we should delete the legacy `setup.py` and use purely a configuration file instead.

I asked ChatGPT to convert this `setup.py` file into that type of system, and it spit this out:

1) In `pyproject.toml`:

```
[build-system]
requires = ["setuptools>=65", "wheel"]
build-backend = "setuptools.build_meta"
```

2) `setup.cfg` in the project root next to `pyproject.toml`

```
[metadata]
name = rgbmatrix
version = 0.0.1
author = Christoph Friedrich
author_email = christoph.friedrich@vonaffenfels.de
classifiers =
    Development Status :: 3 - Alpha

[options]
packages = find:

[options.packages.find]
include = rgbmatrix

[options.ext_package]
ext_package = rgbmatrix

[options.ext_modules]
# empty on purpose; we define modules below

[build_ext]
# nothing here

[extension:core]
sources =
    rgbmatrix/core.cpp
    rgbmatrix/shims/pillow.c
include_dirs =
    ../../include
    rgbmatrix/shims
library_dirs =
    ../../lib
libraries =
    rgbmatrix
extra_compile_args =
    -O3
    -Wall
language = c++

[extension:graphics]
sources =
    rgbmatrix/graphics.cpp
include_dirs =
    ../../include
library_dirs =
    ../../lib
libraries =
    rgbmatrix
extra_compile_args =
    -O3
    -Wall
language = c++

```

3) Delete `setup.py`

4) Build & install `pip install .`
  (or build wheels) `python -m build`

Do see any problems with this?

Maybe it's just cleaner and doesn't actually change the execution path of what happens haha

ty-porter:
Given that this is PR is addressing that no one can build the project at the moment, this should be discussed as a separate PR.

marcmerlin:
is it possible to add a test after this PR to catch basic "this won't work at all" as a test before a PR can be submitted?

PR #1837, opened 2025-12-11
Add interfaces in c# bindings to make testing easier

UniqueDaniel:
Main objective here is to add interfaces to the bindings to allow a user to test their code without needing to go to the matrix. I have been doing this in my projects for some time as I dev in windows then publish to rpi, writing tests against a fake interface makes life a lot easier. 

While testing I also ununified the examples to make configuring for unique test rigs much easier.


bindings/c#/examples/Rotating3DCube/Program.cs
```diff
@@ -7,7 +7,17 @@
...
+// Set this to apropriate settings for your matrix
+var options = new RGBLedMatrixOptions()
+{
+    Brightness= 80,
+    Cols= 64,
+    Rows= 64,
+    GpioSlowdown= 4
+};
+
+using var leds = new RGBLedMatrix(options);
```
gingters: Everywhere else it's `matrix`, not `leds`. When unifying the initialization in the samples, it would make sense to refactor that variable naming here too ;)

bindings/c#/RGBLedFont.cs
```diff
@@ -12,7 +12,18 @@ public class RGBLedFont : IDisposable
...
     /// <param name="bdfFontPath">The path to the BDF file to load.</param>
-    public RGBLedFont(string bdfFontPath) => _font = load_font(bdfFontPath);
+    /// <exception cref="ArgumentException" >Thrown when <paramref name="bdfFontPath"/> is null or empty.</exception>
+    /// <exception cref="FileNotFoundException">Thrown when the specified file does not exist.</exception>"
+    public RGBLedFont(string bdfFontPath)
+    {
+        if (string.IsNullOrEmpty(bdfFontPath))
+            throw new ArgumentException("Font path cannot be null or empty.", nameof(bdfFontPath));
+
+        if (File.Exists(bdfFontPath) == false)
```
smdn: Is there any reason this cannot be `!File.Exists(bdfFontPath)`?

smdn:
It is difficult to agree to adding interfaces only for testing without being provided with actual test code.

The wrapper class API is not yet stable and may change in future work.
Changing `IRGBLedCanvas` or `IRGBLedMatrix` each time such a change occurs is not considered good implementation.
Generally, changes to the `interface` are considered breaking changes and require users to modify their code.
Therefore, even if we decide to add them, it would be better to add them as `internal` for now, rather than `public`.

The following is a proposal for the test implementation.
If you are considering switching implementations for testing, should `IRGBLedCanvas` and `IRGBLedMatrix` be created via factory interfaces like the following?

```diff
+  public interface IRGBLedCanvasFactory {
+    IRGBLedCanvas Create(...);
+  }

   public class RGBLedMatrix {
+     RGBLedMatrix(IRGBLedCanvasFactory canvasFactory)
+     {
+       rgbLedCanvasFactory = canvasFactory;
+     }

-     public IRGBLedCanvas CreateOffscreenCanvas() => new RGBLedCanvas(led_matrix_create_offscreen_canvas(matrix));
+     public IRGBLedCanvas CreateOffscreenCanvas() => rgbLedCanvasFactory.Create(...);
   }
```

Before adding the interfaces, I think we should discuss and clarify what and how we're testing it.

bindings/c#/examples/BoxesBoxesBoxes/Program.cs
```diff
@@ -1,12 +1,15 @@
...
 {
-    Cols = 64,
-    Rows = 64,
-    GpioSlowdown = 3,
-});
+    Brightness= 80,
+    Cols= 64,
+    Rows= 64,
+    GpioSlowdown= 4
+};
```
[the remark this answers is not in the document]
gingters: As of now, I wouldn't really care about formatting. At a future point in time, I would throw a default .editorconfig with csharpier rules in the project, have that formatted automatically in a huge `refactor: global reformatting` commit, and from there on we simply rely on automatic formatting with running dotnet csharpier every time before committing and live happily forever after 😅

marcmerlin:
@UniqueDaniel  can you please address the changes requested as applicable and confirm the end result is fully tested. 
If it is making things better overall, I'm happy to merge this

marcmerlin:

UniqueDaniel:
It will be soon. I have been in hospital and away from home so not the best places to pick it up.

I will drop the variable checks from the PR as they appear to be controversial for now, and neaten up the examples a little further. 

Not for this PR, but are there any preferred c# code styles people would like to work towards?

smdn:
In my opinion, it would be good to maintain consistency within the existing code.
However, there is still no agreed-upon coding style here, like `.editorconfig`.

For now, we should focus on the actual working code, and style adjustments can come later.

UniqueDaniel:
I removed the interfaces for now, but would like to put in a better implementation at a later date as I find them very useful in my code. The other changes are mostly cleanup as requested.

marcmerlin:
so this is still waiting one change/fix before submit, correct?

marcmerlin:
@gingters  can this be submitted as is, even if not perfect?

marcmerlin:
One last chance to object before I submit this.  I do prefer incremental improvements over nothing waiting for a better tomorrow that may or may not arrive :)
But if this is potentially harmful, please let me know

smdn:
Reviews from me weren’t requested, but I checked the changes.
At least, there aren’t any changes that could be described as 'harmful'.
I think merging this PR could be fine.

However, there is still a non-functional code path remaining in the `SwapOnVsync` method of `RGBLedMatrix.cs`.

Also, personally, I think it would be best to explain why `GpioSlowdown = 4` needs to be specified, either in a comment within the code or in this discussion.

marcmerlin:
@uniquedaniel:  I'm fine with submitting as is unless you want to submit a small update before I do.
Please just reply that you are still around and you're fine with commit as is

UniqueDaniel:
Hi, 

I am still around, but in and out of hospital at the moment.

I would say go ahead with it as is and I will do an additional change to address some of the other points when I have more time.

marcmerlin:
thanks for the reply.

PR #1838, opened 2025-12-11
Update README.md

chrisgilldc:
Python fork header mistakenly included, now removed.
Basic information on installing Python bindings included under Python Support section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,3 @@
-RGB LED Python Package
-======================
-This fork was created to smooth installation of the Python rpi-rgb-led-matrix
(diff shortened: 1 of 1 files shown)


evilC:
@chrisgilldc should it not be moved to the top of the [Python Bindings README](https://github.com/chrisgilldc/rpi-rgb-led-matrix-docfix/tree/master/bindings/python)?

chrisgilldc:
There's some duplication. Why don't we just put a link to the Python Bindings Readme in the main README, under the disclaimer that it's a work in progress?

PR #1842, opened 2025-12-15
Add StackToRow mapper

luciengaitskell:
This allows remapping parallel stands from being vertically stacked to an end-to-end layout.

I have two chains of VMapped panels that are end-to-end to make one wide display, where the Pi is in the middle. I end up with the mapper combo: `V-mapper;Rotate:90;StackToRow:F` (I need a final `;Rotate:180` because of the overall rotation of my display).
I was unable to map this with the existing mappers.

When I initially started experimenting with this, #1478 was not merged yet.


marcmerlin:
First, thanks for submitting this, by default I want to accept new mappers that don't touch other code and help people out.
That said, I'm going to ask you to please look at #1478  and see whether this supports the mapping you wrote. 
Basically my incentive to merge new mappers that can already be achieved with the universal mapper, is kind of low, I hope that makes sense.
If this does solve a situation that cannot currently be handled, then sure, i'm up for it.
Or if you can simply get it to work with the universal mapper, I'm also happy to merge in a README change from you that shows the needed syntax do it with the that mapper

lucg:
@marcmerlin Thank you for the quick response.

> please look at #1478 and see whether this supports the mapping you wrote

I gave it a go with the new mapper, and this mapping config had the same effect:
```
--led-pixel-mapper="Remap:512,64|224,0w|192,0w|160,0w|128,0w|96,0w|64,0w|32,0w|0,0w|256,0e|288,0e|320,0e|352,0e|384,0e|416,0e|448,0e|480,0e"
```

It was relatively straight forward for my display, but the interface is definitely a bit more challenging than the existing VMapper, Rotate, Mirror, etc. mappers. The new remap mapper is quite powerful, and I'm glad it's been added. From my understanding, everything (existing) excluding mirroring can be done with #1478.

Is there still interest in maintaining the simple-to-use mappers? I think a StackToRow mapper could still be useful, especially to avoid having large `Remap` configs for bigger setups.

Thanks!

marcmerlin:
Thanks for checking that the new universal mapper can do it, but your point that for simple mappings, its setup is non trivial is taken.
Given that i've merged yours, but I'm going to ask you to send a followup PR to show usage including some kind of picture or ascii art so that people can see at a glance why it's there
https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/README.md

marcmerlin:
For bonus point, please give syntax example for your new mapper along with picture or ascii art, and then the equivalent syntax for the new universal mapper so people can see both options

PR #1844, opened 2025-12-16
Add StackToRow docs

luciengaitskell:
Add docs for #1842 - `StackToRow` mapper


lucg:
@marcmerlin I added the description of the mapper's modes in the 'forward'/mapping direction (existing virtual layout to new virtual layout), along with an example physical layout and the required StackToRow mapping along with the universal mapping equivalent.

marcmerlin:
Looks great, thank you.

Issue #1828, opened 2025-11-27
Master Bug: C# (used to track C# PRs and issues)


marcmerlin:
Any C# users/devs, please subscribe to this bug an make yourselves owners/assignees

marcmerlin:
@scottparise   @digital-synapse  @gingters  please make yourself co-owner if you can help out with C#

marcmerlin:
resolving https://github.com/hzeller/rpi-rgb-led-matrix/pull/1032 until it can be updated, but listing here so you can review and/or update as needed

marcmerlin:
@ivancmz @davemaster are you knowledgeable enough with C# to help triage C# bugs and recommend what C# patches should get merged?
Another ping for @scottparise @digital-synapse @gingters as my last update may have been lost in the mass cleanup spam.

marcmerlin:
@KirillAldashkin  @UniqueDaniel  @smdn @Michael-Richardson  @digital-synapse 
do we still have a C# person onboard?
If no one can maintain the bindings, I'm tempted to disable/remove the bindings as they aren't useful if they don't work

2 security bugs were found:
https://github.com/hzeller/rpi-rgb-led-matrix/security/dependabot/7

https://github.com/hzeller/rpi-rgb-led-matrix/security/dependabot/8

gingters:
I can't see the PRs from dependabot. I'd consider myself an C# expert. The main issue I would have with contributing is that I only have two panels of the same make to test on.
So depending on the type of PR I wouldn't be able to verify it myself but would need to rely on others to run/test changes with physical access to other / more LED panels. That said, I would love to see the C# bindings staying alive.

ivancmz:
Hi, I don't have much free time right now, but I'll try to contribute what I can.

smdn:
Unfortunately, I am not currently using this C# binding.
However, I hope it will continue to be maintained so that others can use it.

I think Dependabot is probably reporting that .NET 6.0 has reached the end of its lifecycle.
I'll create a PR to fix this.
Could you please share the details of the security bug?

I cannot promise continuous contributions, but I would like to help maintain the binding.

marcmerlin:
@gingters @smdn  let me paste the reports you can't see
Looks like upgrading SixLabors.ImageSharp to 2.1.11 or 3.1.11 or better will fix both.
Is that possible?

https://github.com/hzeller/rpi-rgb-led-matrix/security/dependabot/7
Package
Affected versions
[...]

PR #1846, opened 2025-12-19
Add support for streams to C# PlayGif example

evilC:
Adds support for playing streams (As generated by the C++ `led-image-viewer` util using the `-O` flag) to the C# PlayGif example.  

This code was made using AI, it quite possibly does some dumb stuff. I make this PR mainly to foster discussion so that it can be done right.  

In this PR, I also set `--led-no-drop-privs` in the bindings, because it's not in the `RGBLedMatrixOptions` struct, and I don't see any other way of doing it without passing it to your app when you launch it. Not having this set in a C# app is basically the kiss of death - any call to dotnet packages (eg `System.Collections.Concurrent`) will fail after the matrix has been instantiated.

Some stuff I am really not sure about:

- Should it have made the `Bindings` class public rather than internal?
- Should it have made the `InternalRGBLedMatrixOptions` struct public rather than internal?
- I have no clue whether what it has done in `content-streamer.cc` is sane or not

There is also some other stuff in this PR - maybe does not belong in here, but I'll leave it in for now - again, would welcome discussion on these.
- Modification to `.gitignore` in here to ignore stuff in `.vs` folders (Visual Studio folders)
- A `sync-and-build.ps1` script that facilitates transferring files from a windows PC to a Pi using rsync via WSL, and triggering a rebuild of the library

At some point, I plan on converting this to a library, so that it can be directly consumed in people's apps. I will keep the demo executable, but have it call my library. It probably also needs renaming - "PlayGif" is misleading - it will play GIF, PNG, and streams of either.

### DIFF
--- a/bindings/c#/RGBLedMatrix.cs
+++ b/bindings/c#/RGBLedMatrix.cs
@@ -43,10 +43,16 @@ public class RGBLedMatrix : IDisposable
             // users' commandline.
             // As always, as the _very_ first, we need to provide the
             // program name argv[0].
-            var argv = new string[args.Length + 1];
+
+            // --led-no-drop-privs is also added as there seems to be no other way to do it
+            // other than passing it via command-line, and if you call the bindings from a library,
+            // you cannot do that unless you also pass that argument to your app, which just seems silly.
+            // Without no-drop-privs, dotnet cannot load any libraries, so it seems essential
+            var argv = new string[args.Length + 2];
             argv[0] = args[0];
             argv[1] = $"--led-slowdown-gpio={options.GpioSlowdown}";
-            Array.Copy(args, 1, argv, 2, args.Length - 1);
+            argv[2] = "--led-no-drop-privs";
+            Array.Copy(args, 1, argv, 3, args.Length - 1);
 
             matrix = led_matrix_create_from_options_const_argv(ref opt, argv.Length, argv);
             if (matrix == (IntPtr)0)


evilC:
It's also worth noting that from what I can tell, in the C++ led-image-viewer utility, the play code is identical for both non-stream and stream content (Except non-stream content has to unpack first?), whereas in my implementation, there are two separate playback routines.  
I did have a go at merging them, but have not had any luck.  
Anyone got any idea why this is?  
I also note that the C++ app uses GraphicsMagick, whereas the C# app uses SixLabours.ImageSharp. I wonder if this has anything to do with it?
However, I also note that GraphicsMagic appears dead - seems to have had no updates in 8 years.

evilC:
Set `--led-no-drop-privs` in the bindings, updated OP with notes about this

marcmerlin:
Please get folks from https://github.com/hzeller/rpi-rgb-led-matrix/issues/1828 to review this CL. If one of them approves it, I'll merge it.

smdn:
This PR almost rewrites `examples/PlayGIF`.
However, even if this PR is merged, `examples/PlayGIF` itself does not become unnecessary.

So, I think it would be better to leave `examples/PlayGIF` as-is and instead add a new usage example like `examples/PlayGIFWithStreamReader` that uses `StreamReaderWrapper`.
This would make it clearer what changes are necessary and unnecessary in this PR, and what this change actually achieves.

For the addition of `sync-and-build.ps1` and the `--led-no-drop-privs` part, these are completely separate issues.
I think it's better to remove them for now and split them into separate PRs.

@marcmerlin
This PR also includes changes to the C++ area: `lib/content-streamer.cc`
We'll likely need approval from the C++ area team to determine if these changes are acceptable.

For now, null checks for the arguments are missing, so it seems we need to add them.

smdn:
Notes on context:

The need for `--led-no-drop-privs` appears to originate from this discussion.
https://rpi-rgb-led-matrix.discourse.group/t/c-examples-issues/1157

evilC:
> So, I think it would be better to leave `examples/PlayGIF` as-is

I am inclined to agree.

> For the addition of `sync-and-build.ps1` and the `--led-no-drop-privs` part, these are completely separate issues.
> I think it's better to remove them for now and split them into separate PRs.

Agreed

I also think that the privs issue is unrelated to adding support for streams

I will try and find some time to rework this whole thing, but I kind of have other priorities right now. I guess keep it open for the time being, so that at least people can weigh in on the changes made to the C++ and the bindings

marcmerlin:
@smdn on the modifying C++ bits, I'm willing to review small changes to the core library if needed for external hooks, but I wanted those of you who know and understand C# to first review the core change and get to something you're happy with.
Once you're ok with the change, I'm happy to look at the C++ bit before merging.

marcmerlin:
@evilC  looks like people are otherwise happy with this. Please send over the updated version when you can.

marcmerlin:

evilC:
I do want to merge the stuff in at some point, but it probably won't be until some time in the new year. At this point, I am feverishly trying to get my project working for an event I am going to early next year, and I am paranoid about breaking my code if I go messing with it the bindings.  
Also, when I do do it, I will probably just start a new branch, as I think it will be less hassle.
TL/DR - yeah I think you can probably close it for now

evilC:
I am looking to do a bit of work on this now, and I think I am having a slight change of opinon regarding some of the stuff in PR - namely the tooling.  
@marcmerlin are you averse to merging in changes which include build tasks for VSCode or Visual Studio?  
I am not looking to make it mandatory, but I would rather not keep that kind of stuff solely on my local disk (And possibly lost)  
I would also be looking at maybe adding a README to document how to use the tooling

evilC:
@marcmerlin new branch: https://github.com/evilC/rpi-rgb-led-matrix/tree/playgif-streams  
I added a `.vscode` folder with tooling for VSCode on windows  
See the [README  ](https://github.com/evilC/rpi-rgb-led-matrix/blob/playgif-streams/.vscode/README.md)  
It's pretty sweet. Minimal setup, and then you can just select any example C# app, hit F5 and it syncs it with the Pi, attaches the debugger and you can debug on the PC  
It's currently windows specific, but it could be adapted for linux too - would just need alternate tasks that don't use WSL

marcmerlin:
I asked for a review in https://github.com/hzeller/rpi-rgb-led-matrix/issues/1828
All C# stuff, please link there to get reviews.

PR #1848, opened 2026-01-04
Implement content-streamer C# bindings

evilC:
# About
The primary goal of this PR is to add support for playing streams (As generated by the C++ led-image-viewer util using the -O flag) 
I started from scratch with a new PR and took my time this time, being careful not to alter things that did not need to be altered.  
Again, this is very heavy on the AI, so please review changes to C++ and existing C# bindings to check for sanity. Unlike the last PR though, there are very minimal changes to existing code.  

## Differences from the last PR:  
- Does not change the `Bindings` C# class from internal to public.  
  This turns out to have not been needed. It only happened before because code was trying to access the Bindings not via the wrapper I think.  
- I also did not need to change `InternalRGBLedMatrixOptions` from internal to public
- The Bindings for the content-streamer are now a separate bindings file, so if you do not use them, they are not included  
- Existing PlayGif example remains as it was (Apart from a refactor of how the options are handled in the examples, which IMHO is way better)

## C# Examples
- Add PlayStream example
- All examples now use the same `MatrixOptions` static class  
  This allows you to edit one file, and it affects all examples  
  This is largely responsible for the large number of file changes, but it's only replacing the parameters in the matrix constructor, and adding a reference to the MatrixOptions cs file in the csproj of the example
- Updated existing README for examples to clarify regular (Build on Pi) style usage, and mention new style usage (Build on PC)
- Added new VSCode specific readme in the `.vscode` folder
- FontExample now outputs better instructions  
  Change `font-example.exe` to `FontExample`  
  Mention that you need a BDF font)

## Tooling changes
- Add `.vscode` folder for VSCode specific settings
- Add build tasks to build C++ code on the Pi from within VSCode on Windows  
  (This could conceivably be added for linux too, but it currently relies on WSL)
- Add launch tasks to build C# examples locally on PC, rsync them over to pi, launch and debug them

## C++ API
- Expose methods: Create, Next, Rewind, FileStream Create / Delete  
    These allow playing of streams

## C# Bindings + wrappers
- Add new bindings file with new wrapper for content-streamer functionality
- Old Bindings
  - Expose `Clear` in the RGBLedCanvas wrapper to facilitate clearing the screen
  - Expose `_canvas` in RGBLedCanvas wrapper via the `Handle` property 
    AI says this is needed so that we can write directly to the canvas.  
    The old PR did not need it, because the new bindings were lumped in with the existing ones (And maybe because of the change from internal to public?)
  - `RGBLedMatrix` constructor add argument to always enable "No drop privs" mode before calling `led_matrix_create_from_options_const_argv`   
    Seems impossible to do any other way  
    Not having this set in a C# app is basically the kiss of death - any call to dotnet packages (eg `System.Collections.Concurrent`) will fail otherwise if used after the matrix has been instantiated  
    This means that all the examples now work (Well, I did not test the fonts one, dunno how)

### DIFF
--- a/bindings/c#/ContentStreamer.cs
+++ b/bindings/c#/ContentStreamer.cs
@@ -24,15 +24,6 @@ namespace RPiRgbLEDMatrix
         }
 
         /// <summary>
-        /// Clears the given canvas (sets all pixels to black).
-        /// </summary>
-        /// <param name="frameCanvas">The canvas to clear.</param>
-        public static void ClearCanvas(IntPtr frameCanvas)
-        {
-            ContentStreamerBindings.led_canvas_clear(frameCanvas);
-        }
-
-        /// <summary>
         /// Rewinds the stream to the beginning.
         /// </summary>
         public void Rewind()
--- a/bindings/c#/RGBLedCanvas.cs
+++ b/bindings/c#/RGBLedCanvas.cs
@@ -71,9 +71,12 @@ public class RGBLedCanvas
         led_canvas_subfill(_canvas, x, y, width, height, color.R, color.G, color.B);
 
     /// <summary>
-    /// Cleans the entire canvas.
+    /// Clears the canvas (sets all pixels to black).
     /// </summary>
-    public void Clear() => led_canvas_clear(_canvas);
+    public void Clear()
+    {
+        Bindings.led_canvas_clear(_canvas);
+    }
 
     /// <summary>
     /// Draws a circle of the specified color.


evilC:
I added another commit to change how the exclusion of the examples works in the RPiRgbLEDMatrix csproj file  
With the old style
```
        <Compile Remove="examples\**" />
        <None Remove="examples\**" />
```
I was having problems getting the bindings working as a submodule.
In my project, I had an exclusion:
```
    <Compile Remove="rpi-rgb-led-matrix/bindings/c#/**/*.cs" />
    <None Remove="rpi-rgb-led-matrix/bindings/c#/**/*.cs" />
```
But with the old style from RPiRgbLEDMatrix, I then could not reference my wrapper.
If I commented out my exclusion in my project, I could reference my wrapper, but then I had errors like this:
```
The type 'RGBLedMatrix' in 'D:\Data\Code\Rpi\WearWare\src\rpi-rgb-led-matrix\bindings\c#\RGBLedMatrix.cs' 
conflicts with the imported type 'RGBLedMatrix' in 'RPiRgbLEDMatrix, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
Using the type defined in 'D:\Data\Code\Rpi\WearWare\src\rpi-rgb-led-matrix\bindings\c#\RGBLedMatrix.cs'.
```

In this latest commit, I changed the exclusion in the RPiRgbLEDMatrix project to:
```
        <Compile Remove="examples\**\*.cs" />
        <None Remove="examples\**\*.cs" />
```

I am not sure if I have done the right thing, can anyone advise on this?

evilC:
I moved ClearCanvas from the Content Streamer into the Canvas wrapper, as it logically belongs there - it's not inherently anything to do with streams

smdn:
Note: This PR was recreated from #1846.

smdn:
This PR contains too many changes across too many areas to review as-is.
That makes it difficult to reach consensus with other contributors.

As I mentioned before, it would be best to first focus solely on adding the `StreamPlayer` without making any changes to the other examples.

And it appears you are still making changes.
If you plan to continue making changes, it is still too early to review this PR.
If the purpose of this PR is to illustrate what you want to achieve in the end, please let me know once it is complete.

After that, I recommend breaking all the changes into smaller pieces and recreating them as individual PRs.

evilC:
@smdn done
Split into 3 PRs for now, with a 4th to potentially follow
See my comments here: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1828#issuecomment-3708058432

PR #1850, opened 2026-01-04
Vscode tooling

evilC:
This PR adds tooling for VSCode - primarily to facilitate easily working on the C# bindings and examples  

- Add `.vscode` folder for VSCode specific settings
- Add launch tasks to build C# examples locally on PC, rsync them over to pi, launch and debug them
- Add tasks to sync the C++ API code to the pi and build from within VSCode on Windows  
- Added new VSCode specific readme in the `.vscode` folder
- Updated existing README for examples to clarify regular (Build on Pi) style usage, and mention new style usage (Build on PC)

The above are windows only (Uses WSL), however they could easily be adapted for linux too

### DIFF
--- a/.vscode/README.md
+++ b/.vscode/README.md
@@ -1,0 +1,43 @@
+# Instructions for remote developent on Windows using VSCode  
+These instructions are primarily aimed at those wishing to work on the C# bindings, but they are probably relevant if you wish to work on the C++ too.  
+
+## C++
+### Initial setup
+- Set the environment variables on your local PC:
+  - `RGBLEDMATRIX_REMOTE_MACHINE`: Username and hostname of your Pi, eg `root@DietPi`
+  - `RGBLEDMATRIX_REMOTE_DIR`: The remote folder on your Pi where you wish to store the repo, eg `/root/dev/rpi-rgb-led-matrix`  
+    Ensure this folder exists
+- Ensure Passwordless SSH to the Pi is set up for the account you specified in the environment variable
+- WSL must be installed
+- RSync must be installed in WSL
+
+### Working on the C++
+- Make your changes to the code
+- The following tasks are available:
+  - `Sync C++`: Syncs the local folder with the remote (Does not sync the `bindings` folder)
+  - `Build C++`: Builds the C++ library (Auto-runs `Sync C++`)
+  - `Deploy C++`: Deploys `librgbmatrix.so.1` to `/usr/lib` (Auto-runs `Build C++`)
+
+## C#
+### Initial Setup
+- Do all the setup steps for C++
+- Use the `Deploy C++` task to make sure that the linked library is in place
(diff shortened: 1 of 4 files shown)


rwstca:
I can't recommend this change. It enforces a personal development process and style we don't all follow. It also assumes my build and deploy steps in my personal project are the same as yours (they aren't). What if I use a different IDE? Or same IDE but different syncing, build or deploy process? Why aren't those supported? When will they be? That is the wormhole you open up by starting a precedent of this nature.

Your code looks great and I also appreciate the work you put into this with the help of AI, but I'm not sure you are considering the higher-level downstream architectural impact of decisions.

rwstca:
Well if it's limited to the C# bindings and if it improves people's lives. What's the harm

But let me explain my development and build process for context:

- samba server running on raspberry pi.
- open windows folder direct to raspberry pi
- open project folder in VS code
- save file, samba saves it on the raspberry pi
- terminal open with ssh on raspbery pi
- run a bash script that runs `make` on my display executable
- display executable runs other bash scripts, utils, environment prep

I don't use WSL or Rsync as you are doing here. So this VSCode settings file is an individualized way of building and deploying that may not get used by a majority of devs

evilC:
I have actually made significant improvements since submitting this PR. I now have tasks which build the C++ on my PC (MUCH faster), then build the C#, then copy the .so file into the bin folder of my C#. then rsync it across. This also precludes the need to have the source code on the Pi at all.

> I don't use WSL or Rsync as you are doing here. So this VSCode settings file is an individualized way of building and deploying that may not get used by a majority of devs  
TBH, the main target of a system like I described is for C# people who don't really know C++ and just want to try out the library without a load of faffing around. However, with this new local cross-build system I have set up, I can see significant benefits for people working in C++ too. My latest tasks are also more modular - rather than one task depending on another (So you have to use all of them or none of them), they are all standalone, and then I have wrapper tasks which call various sub-tasks in sequence. That way, you can pick and choose what you want. It would also lend itself to people re-using various bits (eg a wrapper task that swaps out windows specific tasks for linux tasks)

rwstca:
> I have actually made significant improvements since submitting this PR. I now have tasks which build the C++ on my PC (MUCH faster), then build the C#, then copy the .so file into the bin folder of my C#. then rsync it across. This also precludes the need to have the source code on the Pi at all.

Just want to mention I don't think this will work anymore after the recent merge of the `-mtune=native march=native` compiler flags that were merged. Because those compile for the native CPU, which means portability is lost in favour of performance optimization. Which was the goal of flicker reduction. So it likely wont work compiling C++ on a PC and moving it to Raspberry Pi. It has to be compiled on the target machine.

evilC:
> I don't think this will work anymore after the recent merge of the `-mtune=native march=native` compiler flags that were merged.  

I [made a PR](https://github.com/hzeller/rpi-rgb-led-matrix/pull/1857) (Which was merged). If you compile on a PC, it disables those flags

PR #1853, opened 2026-01-10
Allow hot-changing of matrix options

evilC:
Apologies if this is a bit vague - maybe if I describe my use-case, it will clarify where my issue lies.  
In my C# app, on very first start, no RGBLedMatrixOptions are set.  
You then go into the options page and select the options you want (Rows, Cols, RowAddressType etc), and it disposes the matrix and canvas and creates new ones.  
However, this is not taking effect for all options - specifically in my case, RowAddressType.  
If you stop the app and re-start, all fine, the new options take effect, but I would like to be able to make this actually work properly.  
I have done some work with AI on this, and it has managed to fix the problem, but it has changed a bunch of files in the library

Is there any appetite for making this change to the library? I know this PR as it currently stands does not put some code in the appropriate place (eg C bridge code in the .cc files), but it's pointless me spending more time addressing this if there's no appetite to merge this change in, or if people want to do it a different way.  

Out of interest, I notice that it is getting rid of this line, so I am guessing it's known that the current implementation leaves something to be desired?
`static GPIO io; // This static var is a little bit icky.`
  
This is about the best summary that I could get out of the AI that was not "war and peace" type lengths:  
## C++ API Checklist

- **`Framebuffer::ResetGlobals()`**: `static void Framebuffer::ResetGlobals();` — idempotently delete/null `sOutputEnablePulser` and `Framebuffer::row_setter_` and clear framebuffer-level caches so `InitGPIO` can rebuild them.

- **`GPIO::ResetState()`**: `void GPIO::ResetState();` — set `output_bits_ = input_bits_ = reserved_bits_ = 0` and reset `uses_64_bit_` (if present). Do not unmap mmapped registers.

- **Hardware-pulser cleanup**: ensure any PWM/clock registers modified by a hardware `PinPulser` are returned to a safe state (either in pulser destructor or a small `ResetHardwarePulsing()` helper).

- **Optional helper**: `void ResetNativeStateForHotReload();` — convenience that calls the above in the correct order.

**Usage / Ordering**

- Stop/destroy the running `RGBMatrix` instance (join/stop update threads).
- Call `Framebuffer::ResetGlobals()` then `s_global_io.ResetState()` (or `ResetNativeStateForHotReload()`).
- Reinitialize: `s_global_io.Init(new_slowdown)` and `Framebuffer::InitGPIO(...)` or create a new `RGBMatrix`.

**Minimal state that must be cleared**

- `output_bits_`, `input_bits_`, `reserved_bits_`, `uses_64_bit_` (GPIO bookkeeping)  
- `sOutputEnablePulser` (pulser instance)  
- `Framebuffer::row_setter_` (row-address logic and cached `last_row_`)  
- Any PWM/hardware registers if hardware pulser was used

### DIFF
--- a/bindings/c#/RGBLedMatrix.cs
+++ b/bindings/c#/RGBLedMatrix.cs
@@ -72,6 +72,16 @@ public class RGBLedMatrix : IDisposable
     /// <returns>An instance of <see cref="RGBLedCanvas"/> representing the canvas.</returns>
     public RGBLedCanvas CreateOffscreenCanvas() => new(led_matrix_create_offscreen_canvas(matrix));
 
+    [System.Runtime.InteropServices.DllImport("librgbmatrix.so.1")]
+    private static extern void framebuffer_reset_globals();
+
+    /// <summary>
+    /// Reset native framebuffer globals so that GPIO/row-address state will be
+    /// reinitialized on the next matrix creation. Useful when changing
+    /// hardware-mapping critical options at runtime.
+    /// </summary>
+    public static void ResetFramebufferGlobals() => framebuffer_reset_globals();
+
     /// <summary>
     /// Returns a canvas representing the current frame buffer.
     /// </summary>
--- a/lib/framebuffer-internal.h
+++ b/lib/framebuffer-internal.h
@@ -98,6 +98,8 @@ public:
                        int dither_bits,
                        int row_address_type);
   static void InitializePanels(GPIO *io, const char *panel_type, int columns);
+  // Reset internal static globals so InitGPIO() can re-run with new params.
+  static void ResetGlobals();
 
   // Set PWM bits used for output. Default is 11, but if you only deal with
   // simple comic-colors, 1 might be sufficient. Lower require less CPU.
--- a/lib/framebuffer.cc
+++ b/lib/framebuffer.cc
@@ -976,3 +976,26 @@ void Framebuffer::DumpToMatrix(GPIO *io, int pwm_low_bit) {
 }
 }  // namespace internal
 }  // namespace rgb_matrix
+namespace rgb_matrix {
+namespace internal {
+  void Framebuffer::ResetGlobals() {
+    if (sOutputEnablePulser != NULL) {
+      delete sOutputEnablePulser;
+      sOutputEnablePulser = NULL;
+    }
+    if (row_setter_ != NULL) {
+      delete row_setter_;
+      row_setter_ = NULL;
+    }
+  }
+}
+}
+
+extern "C" {
+  void framebuffer_reset_globals() {
+    rgb_matrix::internal::Framebuffer::ResetGlobals();
+    // Also reset global GPIO bookkeeping in led-matrix's static GPIO object
+    extern void ledmatrix_reset_global_gpio();
+    ledmatrix_reset_global_gpio();
+  }
+}
--- a/lib/gpio.cc
+++ b/lib/gpio.cc
@@ -203,6 +203,15 @@ gpio_bits_t GPIO::InitOutputs(gpio_bits_t outputs,
   return outputs;
 }
 
+void GPIO::ResetState() {
+  output_bits_ = 0;
+  input_bits_ = 0;
+  reserved_bits_ = 0;
+#ifdef ENABLE_WIDE_GPIO_COMPUTE_MODULE
+  uses_64_bit_ = false;
+#endif
+}
+
 gpio_bits_t GPIO::RequestInputs(gpio_bits_t inputs) {
   if (s_GPIO_registers == NULL) {
     fprintf(stderr, "Attempt to init inputs but not yet Init()-ialized.\n");
--- a/lib/gpio.h
+++ b/lib/gpio.h
@@ -49,6 +49,12 @@ public:
   // Returns the bits that were available and could be reserved.
   gpio_bits_t RequestInputs(gpio_bits_t inputs);
 
+  // Reset internal bookkeeping about which GPIOs have been claimed.
+  // This does not touch the hardware registers directly; it clears the
+  // tracked masks so subsequent InitOutputs/RequestInputs will reconfigure
+  // GPIO pins as needed.
+  void ResetState();
+
   // Set the bits that are '1' in the output. Leave the rest untouched.
   inline void SetBits(gpio_bits_t value) {
     if (!value) return;
--- a/lib/led-matrix.cc
+++ b/lib/led-matrix.cc
@@ -34,6 +34,18 @@
 #include "framebuffer-internal.h"
 #include "multiplex-mappers-internal.h"
 
+// C wrapper to reset global GPIO bookkeeping from external callers.
+extern "C" void ledmatrix_reset_global_gpio();
+
+// Global GPIO instance used by the library. Previously this was a function-
+// local static inside CreateFromOptions; move it to file-scope so the C
+// wrapper can access and reset it.
+static rgb_matrix::GPIO s_global_io;  // file-scope global
+
+// Implement the C wrapper here so it has file scope and can access
+// the file-scope GPIO instance.
+extern "C" void ledmatrix_reset_global_gpio() { s_global_io.ResetState(); }
+
 // Leave this in here for a while. Setting things from old defines.
 #if defined(ADAFRUIT_RGBMATRIX_HAT)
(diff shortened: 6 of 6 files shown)


evilC:
According to the AI, there's a fair few options which cannot be changed at runtime.  
 
In terms of my app, it's way way quicker to re-initialize the matrix (Less than a second) compared to restarting the whole app (~30 seconds).

> perhaps there was an architectural reason or nuance behind why it was not added or supported

If this code works, then surely not?  

WRT trusting the AI - yeah, I hear ya, but I would make two observations (With proper review of course):
1. We already trusted it to do the content-streamer changes
2. It only changes one existing line:  
`static GPIO io; // This static var is a little bit icky.` to `GPIO &io = s_global_io;`  
Everything else is adding new functions.  
So surely unless that one line is doing something bad, then if you don't use the new functionality, there's no risk surely?

evilC:
> C# is just slow is the real problem. There should be methods / tricks to speed up boot time

To be fair, ~30 seconds is not right, it's more like 10 (I was thinking of the time it takes me to build and rsync across too). It's starting up web servers and such, so that's where a lot of the time goes I think.  

C# seems plenty fast enough. Even on a Pi Zero, I can play a stream and be encoding another stream at the same time, and the playback does not suffer.

> What is the list of unsupported runtime options  

Hard to tell. I realized though that I could set some options that I *don't* need, and then try to hot-change to ones I do need, and work out which ones are not currently supported.  

> Are you sure they're not supported in the underlying C++ framework, or is it just a problem of outdated C# bindings? 

It's possible I guess. I just had a quick look at the code, and it appears that calling `Dispose()` on the matrix ultimately calls [here](https://github.com/hzeller/rpi-rgb-led-matrix/blob/master/lib/led-matrix-c.cc#L193). I am not understanding what `to_matrix` is.  

> Sorry for being a hit of a hardass here. The goal is just to probe your understanding of the code  

When it comes to the underlying C++, that's basically zero. I don't hold it against you - I fully appreciate that hacking around with the internals without understanding them needs proper oversight.

rwstca:
Yeah even with extremely optimized C++ code I agree with you, mine takes up to 10 seconds too because I run build steps and all this environment preparation and hook scripts before I run a display.

It should be easy to determine the unsupported options. You just compare the library C++ option parsing code to the RuntimeOptions object. But in your case you also need to validate the separate. underlying C# bindings code because it also duplicates the LED option parsing and handling. I can point you to the exact files but I think its important (learning-wise) to discover this on your own.

C++ is not that harder to understand than C#, you just need to focus on the elemental logic rather than syntax to read it as pseudo-code.

evilC:
Maybe one day I will have a bash at C++, but I have so far spent something like 200 hours on this project (Not just the code, but the CAD, the 3D printing, etc, etc) and I have not even used the thing in anger yet.  I plan on wearing it to an event this weekend - but frankly I am shitting myself that it's gonna crash and burn.  
To be honest, this feature is dedinitely a "Could have". As I have open sourced, it would be nice to be able to get the software in a state where it's something of an appliance with no real knowledge of coding needed to make one.  
For my own purposes, this feature is not needed at all.

marcmerlin:
@hzeller this does touch the core I know you care about, and I have very limited review time and 0 testing ability (on a dive boat at the moment). Can you please review this?

marcmerlin:
I'm not against this in principle, but I will not reasonably be able to review this for a while if @hzeller can't get to it.
If you can find someone comfortable with the C++ bits (@board707 or someone else) who can review this, I'd be open to merging it

board707:
> If you can find someone comfortable with the C++ bits (@board707 or someone else) who can review this,

Hi
Unfortunately, in such complex projects, it's difficult to account for all the nuances that arise with such changes. In particular, I don't think it's safe ​​to replace a type GPIO variable :
```
static GPIO io;  // This static var is a little bit icky.
```
with a reference to GPIO:
```
  // Use file-scope GPIO instance.
  GPIO &io = s_global_io;
```
 It's strange that the code continued to compile after that. Perhaps this variable isn't used at all, but I'm hard to say for sure. 
It would be better to ask the author about such nuances.

Update: 
Am I correct, @evilC , that the modified code compiles and runs? While waiting for someone more experienced to review your code, I'd recommend testing it more extensively with different option sets.

evilC:
Yeah, it ran and worked. I did not test it extensively, I just changed options which previously required a restart to take, and observed that they took immediately without having to restart the app. I did not run the code for any great length of time though.

hzeller:
It is not quite clear to me why it would be necessary to change these parameters after the matrix has been initialized ?

evilC:
> It is not quite clear to me why it would be necessary to change these parameters after the matrix has been initialized ?

I am writing an application that drives the matrix, and I wish to make it as turn-key as possible.
As it stands, when the user first launches the application, all the matrix options will be at default values (And so likely will not work with their panel).  
If they then set the matrix options in the app to match their panel, they would need to SSH into the pi or reboot it or something to restart the application, because some of the matrix options cannot be changed, even though the app disposes of the matrix instance and re-initializes another one.  
As I said, it's not the end of the world if this does not happen, but it would be nice. However, judging by that "This static var is a little bit icky" comment in the existing code, it seems like that is what is causing this issue in the first place, and should be changed anyway?

marcmerlin:
@hzeller  can you give a yeah or nay on this?

marcmerlin (approved):
I'm going to approve this, hopefully this is useful.

hzeller:
I am still not convinced.

If the consensus is that it actually is useful, it will require some rewiew
and changes.

evilC:
@hzeller You won't hear any objections from me on that front.  
I would also like to take this opportunity to raise a little bit of a tangent discussion, but first a little background info (And apologies if this turns into a wall-o-text)  
My app is solely concerned with playing of streams, via the led-image-viewer (Well, technically the content-streamer API)  
Whilst the content streamer does support playlists, it does not meet my needs for many reasons, the main of which is that say the user wants a playlist containing a picture and two animations:  
They want to display the picture for 5 seconds, play the 1st animation for 5 seconds, and then loop the 2nd animation 3 times.  
I would not want to force the user to pre-build GIF files just for this playlist from their source material such that "1 loop" can be specified and everything plays for the desired duration.  
So far that reason, in my app, each item in the playlist is played with it's own content-viewer API call, one after another.  

My app only ever plays streams, it never plays GIFs or PNGs (So that there is no gap between one item and the next)  

Furthermore, the user may want different options for each playlist item (eg limit to a different refresh for playlist item A and B)  

So for that reason, when the user imports an item, they have the opportunity to specify matrix options for that item, which gets stored in a JSON file in the library alongside the stream.  

Also, in terms of the playback, currently I instantiate a matrix (And canvas) instance once, and this is re-used for every item in the playlist.  

Now here's the crux of this post:  
When a playlist item is played, the content-streamer is passed the filename of the stream, which I assume has certain things in the header which reflect the matrix options that were passed to the led-image-viewer at stream creation time.  
The content streamer is then used to get "frames", which are passed to the matrix (Which has it's own options that were passed to it when it was instantiated)  
Some of the options that the matrix was instantiated with will be utterly ignored (eg brightness - that is baked into the stream), some need to match (eg rows, cols), and some will be observed (I am guessing limit-refresh falls into this category?)  

I did do some [initial tests](https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/utils#stream-notes) to work out which options are baked into the stream, but I found it kind of hard going.  
I was thinking of writing some code that took a GIF, ran it through the stream conversion process with no options, saved it as a reference, and then for each option, re-convert the same GIF with that option set to a non-default value and do a file compare to see if the resultant stream was identical or not, but I wonder if you could maybe shed any light on this?

Martin12350:
If I can join, I think it would be really useful in tweaking parameters to avoid flickering. In my case, I found out if I just dispose of the Matrix and reinit with new options, sometimes it works, but sometimes it crashes with a segmentation fault. There is probably an issue with my C# code and how fast I am reinitializing it, and some bigger delay could fix it, but with this "restart" function, there would be no such need to juggle with dispose and hope the underlying library already managed to release all necessary things. Then it would be just a matter of changing a few parameters and calling reset.

marcmerlin:
honestly, this looks simple enough as is and I can see why it can help some cases, so given that this has been waiting since January already, I'm going to lean on the side of merging this.
Henner, you obviously know the code a lot better than me, if you'd like to send a followup CL to improve/clean this, please do

PR #1854, opened 2026-01-10
Add ability for stream reader to check compatibility of stream before playing

evilC:
The content streamer is currently unable to tell the difference between a stream that has reached the end, and an incompatible stream.  
When we call GetNext(), if the stream is incompatible, it will return false, which will be mis-interpreted as having reached the end of the stream.  
This PR adds functionality to detect if a stream is valid before attempting to play it.

### DIFF
--- a/bindings/c#/ContentStreamer.cs
+++ b/bindings/c#/ContentStreamer.cs
@@ -36,8 +36,7 @@ namespace RPiRgbLEDMatrix
         /// </summary>
         public bool IsCompatible(IntPtr frameCanvas)
         {
-            if (_streamIO == IntPtr.Zero)
-                throw new InvalidOperationException("Stream IO not initialized");
+            ObjectDisposedException.ThrowIf(_streamIO == IntPtr.Zero, GetType());
             if (frameCanvas == IntPtr.Zero)
                 throw new ArgumentNullException(nameof(frameCanvas));
             return Bindings.file_stream_io_is_compatible_with_canvas(_streamIO, frameCanvas);


bindings/c#/examples/StreamPlayer/StreamPlayer.cs
```diff
@@ -18,10 +18,14 @@ public StreamPlayer(RGBLedMatrixOptions options)
...
     /// <param name="streamPath">The path to the stream file.</param>
     /// <exception cref="InvalidOperationException"></exception>
-    public void PlayStream(string streamPath)
+    public bool PlayStream(string streamPath)
     {
         using var contentStreamer = new ContentStreamer(streamPath);
         var canvas = _matrix.CreateOffscreenCanvas();
+        if (!contentStreamer.IsCompatible(canvas.Handle))
+        {
+            return false;
```
smdn: The `PlayStream` method continues execution until interrupted by Ctrl+C. Consequently, it does not return true until the playing is stopped.
Therefore, throwing an exception to indicate that playback could not be started would be more appropriate than using the return value to signal an error in this case.

```cs
if (!contentStreamer.IsCompatible(canvas.Handle))
{
    throw new InvalidDataException("The stream file is not compatible with the current matrix configuration.");
}
```

bindings/c#/ContentStreamer.cs
```diff
@@ -31,6 +31,18 @@ public void Rewind()
             content_stream_reader_rewind(_reader);
         }
 
+        /// <summary>
+        /// Checks whether the underlying stream is compatible with the given canvas.
+        /// </summary>
+        public bool IsCompatible(IntPtr frameCanvas)
+        {
+            if (_streamIO == IntPtr.Zero)
+                throw new InvalidOperationException("Stream IO not initialized");
```
smdn: The `_streamIO` field is guaranteed to be initialized with a non-zero value in the constructor.
And a value of `IntPtr.Zero` indicates that the handle has been released by the `Dispose()` method.
Therefore, in this situation, the appropriate exception to throw is `ObjectDisposedException`.

```diff
- if (_streamIO == IntPtr.Zero)
-     throw new InvalidOperationException("Stream IO not initialized");
+ ObjectDisposedException.ThrowIf(_streamIO == IntPtr.Zero, GetType());
```

bindings/c#/examples/StreamPlayer/StreamPlayer.cs
smdn: The main point of interest lies in the design choice of returning a success or failure status, particularly in the behavior where the method does not return `true` immediately upon success, but only after entering a while loop and continuing until it is explicitly stopped.

When incompatibility is detected, the method returns `false` immediately, which seems fine.
However, when compatibility is detected, the method enters a loop and never returns a result until it is explicitly stopped.
In that case, the returned `true` doesn't carry much significance as a method result.

Therefore, throwing an exception in cases of incompatibility, rather than returning `false`, and simply entering the loop when compatible might offer a clearer design.

On top of that, having a method like `static bool StreamPlayer.IsCompatible(string path)` might be helpful for checking compatibility before starting playback.

smdn: My point was about how the current `PlayStream()` method behaves in the example.

> The API returns true / false immediately. The example does not, because it is set to play the stream indefinitely.

That's exactly what I was trying to highlight.
The fact that the stream plays indefinitely, yet the method appears to potentially return `true`, may seem like an unexpected behavior at first glance.
That said, as this is just an example, strict design consistency isn't required, and I don't think there's any need to continue the discussion here.
I will leave the final decision to you.

> Surely it can't be static, as it needs a canvas and matrix instance?

That is correct. I mistakenly thought it could be static.
What I meant to suggest was that enabling compatibility checks to be performed from outside the class, prior to calling `PlayStream`, might make the example more illustrative.
But if you feel that's unnecessary, then it’s fine to leave it out.

evilC: > enabling compatibility checks to be performed from outside the class, prior to calling `PlayStream`, might make the example more illustrative

Ah, I see where you are coming from now, and I think I agree.
[Edit] In principle. However, this would require instantiating a content streamer to make the check, doing the check, disposing of the content streamer, and then re-instantiating one when we go to play. This is just gonna chew more CPU and slow things down.

smdn: Right.
If `streamPath` were specified in the constructor of `StreamPlayer` rather than in the `PlayStream` method, it would no longer be necessary to instantiate solely for compatibility checks.
However, whether it is worth modifying the design to add `IsCompatible` as a member of `StreamPlayer` is a decision I will leave to your discretion.

evilC: The way I use it in my app is I keep one StreamPlayer instance (With matrix and canvas already initialized), and then re-use it for all the streams that play, so I don't really see that would solve that issue?  
To be fair, I just noticed, I should probably have moved the canvas to being a field that was initialized in the ctor, like I do in my app.  
I mean, of course, we're talking about the example here, but anyway, I would assume if someone were using the StreamPlayer example as a basis, they would probably do the same.
Anyway, gonna resolve this, as it's not stopping the merge from proceeding.

Issue #1754, opened 2025-01-21
PREEMPT_RT kernel option


gregtap:
👋🏻 Hello Led lovers,

I did not see any conversation about tunning Kernel for better performance i.e increased refresh rate or less flicker interruption.

Is there some proven benefits around `PREEMPT_RT` kernel flag?

hzeller:
There might. I've tried it a couple of years back, but the result was inconclusive with various versions of the Raspberry Pi. It was a pain to compile the kernel back then so given the results I didn't bother to explore further.

Having said that, that was at least 6-7 years ago, so if someone wants to work on newer experiments, that would be great.

kjkent:
@marcmerlin From some cursory googling, it doesn't seem so, though it looks like ~ 2017 they experimented with offering it via the .txt config files: https://dietpi.com/forum/t/rpi-3-v157-rt-kernel-not-updated/1592/2.

For tweaks for rt-sensitive workloads, this is a 5-year-old post (https://dietpi.com/forum/t/dietpi-for-real-time-application/4373) but DietPi project lead Michaelng details scheduler changes that may benefit, as well as adjusting process niceness...

> Little article about what is possible: https://www.redhat.com/sysadmin/real-time-kernel
>
>However, lets see what we can do. Nice -20 is already good. But the CPU scheduling policy for the process is probably not optimal: https://manpages.debian.org/stretch/manpages/sched.7.en.html
>For real-time processes switch to FIFO (first in-first out) or RR (round-robin). If you run it as systemd service, dietpi-services can change it on service-level. To be true right now I don’t know the command to switch this outside systemd, but should be not difficult to find. The CPU priority can then be set as kinda replacement for nice.
>
>Does your process as well read and/or write to disk? In this case, the same as for CPU scheduler applies for I/O scheduler: https://manpages.debian.org/stretch/util-linux/ionice.1.en.html
Ah, its the ionice command. Use realtime for scheduling class and the I/O scheduling priority, although AFAIK that is raised automatically together with nice.
>

...Perhaps there are even better options now.

[...]

PR #1855, opened 2026-01-11
Compiler performance optimization in pursuit of flicker reduction

rwstca:
Adds the following gcc compiler optimization flags to all applicable compilation stages within the repo (lib, examples, utils, python/bindings):
- `-march=native`
- `-mtune=native`
- `-flto=2`

The benefits of this are to generate optimized performant machine code that is targeted specifically towards the native CPU instruction set for the Raspberry Pi. This helps reduce flicker, increase refresh rate and general performance on resource-limited hardware like the Raspberry Pi Zero 2.

It would require more personal time from me that I currently don't have to generate performance benchmarking comparisons within this PR, but the flags themselves are well-documented within the gcc ecosystem and are reliable. I have tested this work on my personal project and it is working as intended.

On a general example (according to AI), machine code generated with these flags such as an intensive computational loop are 2-3x faster when benchmarked with a simple start/stop timer.

### DIFF
--- a/lib/framebuffer.cc
+++ b/lib/framebuffer.cc
@@ -19,6 +19,7 @@
 
 #include "framebuffer-internal.h"
 
+#include <algorithm>
 #include <assert.h>
 #include <ctype.h>
 #include <math.h>
@@ -52,8 +53,10 @@ PixelDesignator *PixelDesignatorMap::get(int x, int y) {
 
 PixelDesignatorMap::PixelDesignatorMap(int width, int height,
                                        const PixelDesignator &fill_bits)
-  : width_(width), height_(height), fill_bits_(fill_bits),
-    buffer_(new PixelDesignator[width * height]) {
+  : width_(std::max(width, 1)),
+    height_(std::max(height, 1)), 
+    fill_bits_(fill_bits),
+    buffer_(new PixelDesignator[std::max(width, 1) * std::max(height, 1)]) {
 }
 
 PixelDesignatorMap::~PixelDesignatorMap() {


evilC:
I will try and do some tests with this - I am guessing I could just use the `led-image-viewer` util with the `--led-show-refresh` parameter set, and expect to see a higher refresh rate?

evilC:
I just did a test with the led-image-viewer and I see no difference  
Here are the parameters that I am using:

Basic - 
`./led-image-viewer --led-rows=64 --led-cols=128 --led-slowdown-gpio=2 --led-parallel=2 --led-row-addr-type=5 --led-no-drop-privs --led-show-refresh /root/rotating_cubes.gif`  
I get about 97hz, 76% usage of core 4 for both the original code and your code  


Trying to push speed
`./led-image-viewer --led-rows=64 --led-cols=128 --led-slowdown-gpio=2 --led-parallel=2 --led-row-addr-type=5 --led-no-drop-privs --led-show-refresh --led-pwm-dither-bits=1 --led-pwm-lsb-nanoseconds=50 --led-pwm-bits=7 /root/rotating_cubes.gif`  
I get about 366hz, 100% usage of core 4 for both the original code and your code  

Tested on an Rpi Zero 2W  
Attached is the GIF I used (I used others too, but this one changes a lot of pixels in pretty much every frame, so I thought it would be a good test). I'll attach my 128x128 version and the full size versiom pre-crop / resize  
whyte-c
rotating_cubes

Just to be sure, I double-checked I was using the right code. Looks like it:
```
root@DietPi:~/dev/rwstca-fork/rpi-rgb-led-matrix# more examples-api-use/Makefile
CFLAGS=-Wall -O3 -march=native -mtune=native -flto=2 -g -Wextra -Wno-unused-parameter
```

rwstca:
I think it's awesome you took the time to test this out further and I appreciate it. Let me just verify some points here. 

- This change is not a god-fix, but rather one element of a wider holistic plan to tackle flickering.  (more info here: https://github.com/hzeller/rpi-rgb-led-matrix/issues/1754#issuecomment-3738460058)
- The compiler flag here will not result in a 1:1 improvement in raw FPS speed, as there is still timing requirements by the external LED panels and rgbmatrix library internal clocking cycles
- The end-goal is not the refresh rates, it's flicker reduction - so we should not use that as the main scoring metric for the PR.
- The flickering can feel subjective at times. At night when I'm tired I blink more so it can seem iffy day-to-day.
- ~It looks like you are a C# bindings user, so I'm not fully sure your final C# executable is reaping the benefits of the C++ machine code compiler optimization.~ (Nvm, you are using led-image-viewer which is a C++ executable)
- Support is currently for C++ and Python bindings only (I am unable to support C# as I do not use it or have the proper environment)

evilC:
No problem happy to give feedback.  
> Nvm, you are using led-image-viewer which is a C++ executable  

I did the tests using the image-viewer, but in my app I call the content-streamer API via the new C# bindings that I wrote for it.  
I only use the led-image-viewer executable in my app to create the streams (I did look at exposing stream creation via an API, but it got very, very messy so I gave up)  
If you wanna have a play, it should be *somewhat* turn-key (Basically set some env vars on your PC, create a folder for the app on the pi, drop some GIFs in the `incoming` folder on the Pi and hit F5 in vs code). The source is here: https://github.com/evilC/WearWare

Issue #1852, opened 2026-01-05
Deprecated warning when building library


evilC:
I am seeing this when building:

```
demo-main.cc:1010:27: warning: ‘void std::random_shuffle(_RAIter, _RAIter, _Generator&&) [with _RAIter = GeneticColors::citizen*; _Generator = int (&)(int)]’ is deprecated: use 'std::shuffle' instead [-Wdeprecated-declarations]
 1010 |       std::random_shuffle (children_, children_ + popSize_, rnd);
      |       ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/14/algorithm:61,
                 from demo-main.cc:23:
/usr/include/c++/14/bits/stl_algo.h:4531:5: note: declared here
 4531 |     random_shuffle(_RandomAccessIterator __first, _RandomAccessIterator __last,
      |     ^~~~~~~~~~~~~~
```
I know zilch about C++, but I can try to fix it using AI
Unless anyone else wants to have a go?
Ot just leave it?

marcmerlin:
I'm travelling for 2 weeks and can't check easily, but does the warning go away and the code still work if you
random_shuffle => std:shuffle?

evilC:
> I'm travelling for 2 weeks and can't check easily, but does the warning go away and the code still work if you random_shuffle => std:shuffle?

I changed
```
    while (!interrupt_received) {
      swap();
      sort();
      mate();
      std::random_shuffle (children_, children_ + popSize_, rnd);
```

to
```
    while (!interrupt_received) {
      swap();
      sort();
      mate();
      std::shuffle (children_, children_ + popSize_, rnd);
```

Then I get
```
make[1]: Entering directory '/root/dev/rpi-rgb-led-matrix/examples-api-use'
g++ -I../include -Wall -O3 -g -Wextra -Wno-unused-parameter -c -o demo-main.o demo-main.cc
In file included from /usr/include/c++/14/algorithm:61,
                 from demo-main.cc:23:
/usr/include/c++/14/bits/stl_algo.h: In instantiation of ‘void std::shuffle(_RAIter, _RAIter, _UGenerator&&) [with _RAIter = GeneticColors::citizen*; _UGenerator = int (&)(int)]’:        
demo-main.cc:1010:20:   required from here
 1010 |       std::shuffle (children_, children_ + popSize_, rnd);
[...]

PR #1856, opened 2026-01-12
Fix warning on build

evilC:
PR to fix issue #1852


rwstca:
This looks good to me. I'm also sick of the warnings during my build stage. Let's get rid of them. Thanks for the work on this!

Also just wanted to mention, the instantiation of `std::mt19937 rng_;` is ideal. It initializes a random number generator engine which is more efficient than the previous implementation of `rand()`.

PR #1862, opened 2026-01-27
Add support for Raspberry Pi CM4S model

calcinai:
Adds detecting of revision `a03150` etc for CM4S. Without this, it uses the wrong base register address as the CM4S uses the BCM2711

### DIFF
--- a/lib/gpio.cc
+++ b/lib/gpio.cc
@@ -321,6 +321,7 @@ static RaspberryPiModel DetermineRaspberryModel() {
   case 0x11: /* Pi 4 */
   case 0x13: /* Pi 400 */
   case 0x14: /* CM4 */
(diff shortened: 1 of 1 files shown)


marcmerlin:
Thank you for that.  Out of curiosity, are you still using this and just getting 3 parallel outputs, or are you getting more?
In your project, how is the CM4 a better choice than a regular RPi4?

calcinai:
@marcmerlin Using this in a couple of places on a custom carrier, main reason is not for the extra chains but for the ability to directly expose some of the other buses not available otherwise. The boards expose 4/5 chains but we're only using two or three.

We currently also compile with the WIDE_GPIO, where we should actually change that to allow the CM mapper without assuming we'd consume all 6 chains.

PR #1887, opened 2026-04-27
RT Kernel Documentation Update

kingdo9:
RT Kernel Documentation Update

### DIFF
--- a/RT-kernel/README.md
+++ b/RT-kernel/README.md
@@ -27,12 +27,14 @@ isolcpus=domain,managed_irq,2,3 nohz_full=2,3 rcu_nocbs=2,3 irqaffinity=0,1 idle
 
 Full image with Hzeller library installed + RT Kernel.
 
-<a href="https://mega.nz/file/GAMEyDzS#9MvAs_ffPo4cROvovjziP0neAxOb1h7nzIK7DXxaYHo">Raspberry Pi 4 Lite Trixie 64-Bit - RT Kernel Preinstalled image</a>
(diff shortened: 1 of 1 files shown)


marcmerlin (approved):

marcmerlin:
@kingdo9  I understand that you found other kernel changes make flickering go away, but I'm not sure RT kernels are useless. Any kernel driver can decide to grab locks and take focus away from our app.  Having RT should definitely help for this not to happen.
So since you went through the effort of making an RT kernel and your suggestions do include another kernel anyway, I so recommend we keep RT in unless it's been found to make other things worse

marcmerlin:
From what you said, RT can indeed override some drivers like USB ones (you said ethernet or serial), but is there a way to keep RT in and only turn it on for people who want to benefit from RT and don't have those drivers that don't do as well with it?

kingdo9:
@marcmerlin 

Not possible to do selective RT.
The main fixes that made the actual difference are kept in the new optimized kernel.
I never saw any improvement with the RT flag alone.

I think because our app is constantly running, it holds priority regardless especially as we use
taskset -c 2 chrt -f 99
RT could be dealing with unrelated processes / interrupts at a higher priority which we wouldn't want anyway?

My displays are running exactly the same. I'm very fussy when it comes to it, if I saw any degredation I wouldn't go this way
When you get time maybe you could compare? Didn't you see the different previously? Can you compare on this?

There was someone in the discourse forum who also immediately confirmed it resolved their flickering issue. I could try reach out to them also.

https://github.com/hzeller/rpi-rgb-led-matrix/pull/1889

marcmerlin:
thanks for the details and the work, I'll do my best to look at it, but it will likely be a few more weeks before I'm home and I can do that

PR #1889, opened 2026-04-29
Optimized Kernel - RT Kernel Not Required

kingdo9:
Confirmed Real Time kernel with Zero2W and Pi 3 was increasing CPU usage when using USB devices.

The RT specific parameter was found not to be reducing flicker, but it was the other options associated with it which I have kept and make the flicker reduction difference.
https://github.com/kingdo9/rpi-rgb-led-matrix_pwm_experiment/tree/Optimized_Kernel_Documentation_Update/optimized-kernel

You can read about the bug here
https://forums.raspberrypi.com/viewtopic.php?t=159170

"I have submitted the matter to the RT_PREEMPT list, and the developers suggest that there is a bug in the FIQ USB implementation, which is only made more evident by the usage of the RT patch.
You may try and disable the FIQ in the command line. If the freezes disappear also in your cases, then we may confirm the bug and submit the matter to rpi3 kernel people."

---

Another user who was using the RT kernel and a USB serial adapter encountered crashing
https://rpi-rgb-led-matrix.discourse.group/t/crashing-rt-kernel-for-pizero2w-confirmation-help-requested/1218

If you get time to test, try a Zero2W or Pi 3 with a USB Ethernet adapter and look at the CPU usage when using RT kernel.

This pull request keeps the features of the flicker reduction but without RT parameter. From my testing the performance is the same. @marcmerlin when you get time are you able to confirm the same performance with this optimized kernel?

RT does not make the actual difference, it is the below options I have kept in the optimized kernel.

isolcpus=domain,managed_irq,2,3 nohz_full=2,3 rcu_nocbs=2,3 irqaffinity=0,1 idle=poll
Timer tick handling > Full dynticks system (tickless)
And additional patches for Zero 2W and 3 to move more interrupts off core 2.3

---

Example CPU usage - Old RT Kernel

rt_kernel<br>

Example - New Optimized Kernel without RT

optimized_kernel


marcmerlin:
thanks for confirming. I am not home and won't have time home for a while, so it cannot test this for quite a while, sorry.
I will when I get home, but for now I have your RT kernel + patches from earlier running on my home frames 
Please get at least one more person who tried your new kernel and had issues with the RT one to comment on this bug

PR #1892, opened 2026-06-06
Pi5 - Improvements, additional led-row-addr-type support

kingdo9:


marcmerlin:
Pasting from @ty-porter:
Not sure if there's a better place for this, but #1892 seems to break the CLI API contract (including for Python and C# bindings) for no good reason.

--led-rp1-pio replaces --led-rp1-rio with no deprecation warning.

The functionality appears identical in terms of selecting PIO/RIO so I'm not sure why we didn't simply swap the default value for that flag instead. It's not the end of the world, but I think we should be careful about changes like this.

marcmerlin:
@ty-porter  thanks for bringing that up, I missed that in my review.
Am I correct that this is a new flag that's only been around for a few weeks and few people should be using it, yet?
If so, it doesn't sound terrible enough for me to revert the the CL, but on principle I also agree with you it's not great to break flags.

@kingdo9  can you review this comment/concern, and provide a small tweak/fix for that if possible?

kingdo9:
> --led-rp1-pio

I wanted rio to be default as it was the best performing. 
It was not ideal to swap the flag name but there were issues with how the options were used across the library, mainly the c# bindings. 

instead of assigning led-rp1-pio=0 to enable rio, - Having a negative integer to enable a feature wasn't working well with the c# bindings. 

So I needed to change --led-rp1-rio  to replace with --led-rp1-pio

Pi5 RIO mode is what it uses by default as it offers the best performance. Being default no flag should be needed.
To use PIO mode you would use --led-rp1-pio=1

apologies for the change, but the rio mode was a later finding during the whole process and was the better performer.

it would be messy to make it backward compatible, hopefully this design I see as final now. the downstream projects would be best updating to this new flag.

marcmerlin:
Given that it was a recent commit and addition, I'm willing to let things be as is this time around. It's of course not ideal, but it should not affect too many users given how recent it was

ty-porter:
Thank you both for sharing your reasoning.

marcmerlin:
And thanks for flagging this @ty-porter , if it hadn't been for the bits that made this an exception this one time, I would have wanted to revert/fix this.

PR #1893, opened 2026-06-07
Optimized Kernel - Migrate prebuilt kernel to another file host.

kingdo9:
As per @hzeller request
https://github.com/hzeller/rpi-rgb-led-matrix/pull/1890#issuecomment-4641546640

Would this need removing from the whole history? I believe using git-filter-repo but would require a force?
I'm not too familiar on that but this PR removes the file and links to another file host.


hzeller:
Can you also add a documentation how to patch, configure and compile the kernel ? I would be interested in that, as it helps understanding all the details.

kingdo9:
> Can you also add a documentation how to patch, configure and compile the kernel ? I would be interested in that, as it helps understanding all the details.


The guide is within the readme. I advise all the options, precompiled or compile your own.

https://github.com/hzeller/rpi-rgb-led-matrix/tree/master/optimized-kernel#option-3---compile-and-install-your-own-raspberry-optimized-kernel

hzeller:
Ah, thanks, I missed that.
Do you consider getting the Raspberry Pi foundation interested in hosting the kernel images ? I am not very comfortable sending users to a random unverified kernel binary.

kingdo9:
> Ah, thanks, I missed that. Do you consider getting the Raspberry Pi foundation interested in hosting the kernel images ? I am not very comfortable sending users to a random unverified kernel binary.

I wouldn't know the process of that. Unless yourself compiles it so you can verify the build.
or just keep the instructions only to compile own.
One kernel for all the main Pi versions which helps.

hzeller:
Alright, let's see what we can do. Thanks for working on figuring out the improved kernel situation!

kingdo9:
> Alright, let's see what we can do. Thanks for working on figuring out the improved kernel situation!

To confirm that kernel was compiled and uploaded by myself. Not sure if you knew that :) but even if still unsure, I understand.

marcmerlin:
Howdy, I had a quick offline chat with @hzeller  but to sync up with everyone.
1) I believe I did actually ask @kingdo9  to contribute the kernel he had built, because I know how much of a pain it is to cross build a kernel, including very very slow (if at all) if you are trying to build it on a Pi because you don't have linux otherwise.  This is another way to say that my call was very few users of this lib would reasonably be able to build this kernel
2) I also personally verified that the kernel fixes display flickering issues I've had forever on almost all my displays and it finally fixes an issue that had been unfixed all these years. I wish the basic RT kernel from upstream would have done the job, but it did not when people tried.
3) I understand shipping a binary kernel comes with trust issues, but I balanced that against point #1 above and the fact that it's opt in if people cared, as well as if security is important, they can build their own with the instructions that were already given.
4) moving the link to another hosting provider works for me, as long as it won't randomly disappear in a few months or a year.  Reading google, it seems that mega.nz downloads can be deleted after 90 days of inactivity, which is not great.

On the plus side, the files can be retrieved from git history if we never need new versions. 
Otherwise, my suggestion is
1) the temporary move to mega.nz is ok but won't work long term
2) @hzeller trusts himself hopefully :)  and we hopefully trust him too, so if he builds those kernels himself and merge the binaries back in, this should be good enough.
3) he mentioned some fancy auto build option done by github but my take is that's it's way too much setup time for something that will only very occasionally need to be updated. I feel like a manual build and merge is enough on time spent vs the need at hand.

PR #1896, opened 2026-07-05
[Bindings/C#] Upgrade TargetFramework(s) to .NET 10.0

smdn:
This PR updates the `TargetFramework(s)` in each project file from `net8.0` to `net10.0`.

.NET 10.0 LTS packages are already available on major platforms where .NET can be installed, such as Ubuntu (`dotnet10`).
In addition, .NET 8.0 LTS will reach EOL this November.
For these reasons, updating to .NET 10.0 at this point is preferable.

This PR updates the target .NET versions as follows:

### `bindings/c#/RPiRgbLEDMatrix.csproj`
`TargetFrameworks` is updated by adding `net10.0` alongside the existing `net8.0`.

If this project file is referenced directly for builds, removing `net8.0` would cause build failures. Therefore, this project should retain `net8.0` for the time being and remove it once it reaches EOL.

### other *.csproj under `bindings/c#/examples`
`TargetFramework` is changed from `net8.0` to `net10.0`.

These project files are unlikely to be referenced by other projects, so simply switching to `net10.0` should have minimal impact.

For reference, the update from .NET 6.0 to .NET 8.0 was made in #1832.

### DIFF
--- a/bindings/c#/README.md
+++ b/bindings/c#/README.md
@@ -8,7 +8,7 @@ To build the C# wrapper for the RGB Matrix C library you need to first have __.N
 
 ### Install .NET SDK
 
-`sudo apt install dotnet8` should work in most cases.  
+`sudo apt install dotnet10` should work in most cases.
 For some old distributions, read [docs](https://learn.microsoft.com/dotnet/core/install/linux)
 Then, in the `bindings/c#` directory type: `dotnet build`
 To run the example applications in the c#\examples\EXAMPLE folder: `sudo dotnet run`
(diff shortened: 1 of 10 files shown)


marcmerlin:
Thanks.  For this and any other C# changes, please get a reviewer from https://github.com/hzeller/rpi-rgb-led-matrix/issues/1828 
to give you a +1 and I'll merge it
Apologies for the delay in answering

smdn:
@ivancmz, thank you for your review! https://github.com/hzeller/rpi-rgb-led-matrix/issues/1828#issuecomment-5031619816

smdn:
Thank you all for your work!

evilC:
@marcmerlin Sorry for being late to the party here - just to confirm that I updated my code to this version, and all is good.

smdn:
Thank you for testing this change using the working code!

